PR #2, opened 2024-03-06
Adding indextool files into the repository

esimpsons3ti:
Adding the following files into the rms-pds4indextools directory:
- pds4_create_xml_index.py
- requirements.txt
- sample_elements.txt
- pds4indextools.ini (the default configuration file)


pds-admin (dismissed):
- The `requirements.txt` file needs to be at the top level of the repo
- `sample_elements.txt` should probably go under `test_files` if you want to keep it at all. No need for the placeholder file. Although if this isn't actually a file being used by your test suite, you could create a top-level folder called `samples` or something for your own use. Then if you have no tests, you don't need to create a `test_files` directory for now.
- If you don't have any tests, you don't need to create the `tests` directory.
- See https://github.com/SETI/rms-devenv/blob/main/github_administration/02_config_python_package.md for more information on configuring the directory and setting it up as a package (you can do this as a separate pull if you want). Note that since we aren't expecting people to import this package, just run it from the command line, you don't need the `__init__.py` file.
- You can go ahead add the top-level README.md file. You can copy it from any of our other repos, replace the old repo name with `rms-pds4indextools` everywhere, and delete any random other text in the README that isn't relevant.

rfrenchseti (approved):
This is a fine start. We'll figure out the packaging later.

Just to be clear I was talking about grabbing a README.md from another repo, one that has all of those colored badges already in it.

PR #9, opened 2024-03-29
Introducing new features and changing XPath numbering

esimpsons3ti:
This pull request does the following:

- All headers in the resulting index file are now `XPaths`
- `--xpaths` is now `--disambiguate-xpaths`. Any unique `XPaths` will be changed into a shortened version.
- Information within the index file is no longer missed due to duplicate generalized `XPaths`
- New function: `grab_elements()`, for your `XPath` comparison needs. Returns a tuple of the element names in the order they appear in the `XPath`.
- covert_xpath_to_tag has been removed, since it's functionality is no longer needed.
- New function: `renumber_xpaths()`, which replaces the numbering in the XPaths to represent their occurrence rather than their hierarchy within the label file.

### DIFF
--- a/pds4indextools/pds4_create_xml_index.py
+++ b/pds4indextools/pds4_create_xml_index.py
@@ -440,6 +440,7 @@ def main():
         if args.disambiguate_xpaths:
                 elements = ()
                 xpath_elements = []
+                tags = []
                 for key in list(xml_results.keys()):
                     xpath_elements.append(grab_elements(key))
                 
@@ -447,8 +448,10 @@ def main():
 
                 for key in list(xml_results.keys()):
                     elements = grab_elements(key)
-                    if elements not in duplicates:
-                        value = elements[-1]
+                    tag = elements[-1]
+                    if elements not in duplicates and elements[-1] not in tags:
+                        value = tag
+                        tags.append(tag)
                     else:
                         value = key
                     xml_results[value] = xml_results.pop(key)
--- a/pds4indextools/pds4_create_xml_index.py
+++ b/pds4indextools/pds4_create_xml_index.py
@@ -1,3 +1,47 @@
+"""
+XML Bundle Scraper
+This script scrapes XML files within specified directories, extracts information from
+user-defined XML elements, and generates a CSV index file. The script provides options
+for customizing the extraction process, such as specifying XPath headers, limiting
+search levels, and selecting elements to scrape.
+Usage:
+    python xml_bundle_scraper.py <directorypath> <pattern>
+        [--elements-file ELEMENTS_FILE]
+        [--disambiguate-xpaths]
+        [--output-file OUTPUT_FILE]
+        [--verbose]
+        [--sort-by SORT_BY] 
+        [--clean-header-field-names]
+        [--extra-file-info EXTRA_FILE_INFO]
+        [--config-file CONFIG_FILE]
+Arguments:
+    directorypath        The path to the directory containing the bundle to scrape.
+    pattern              The glob pattern(s) (which may include wildcards like *, ?,
+                         and **) for the files you wish to index. Multiple patterns
+                         may be specified separated by spaces. Surround each pattern
+                         with quotes.
+    --elements-file ELEMENTS_FILE
+                         Optional text file containing elements to scrape.
+    --disambiguate-xpaths
+                         Replace unique XPath headers with shortened versions.
+    --output-file OUTPUT_FILE
+                         The output path and filename for the resulting index file.
+    --verbose            Activate verbose printed statements during runtime.
+    --sort-by SORT_BY    Sort the index file by a chosen set of columns.
+    --clean-header-field-names
+                         Replace the ":" and "/" with Windows-friendly characters.
+    --extra-file-info EXTRA_FILE_INFO  
+                         Add additional column(s) to the index file containing file or
+                         bundle information. Possible values are: "LID", "filename",
+                         "filepath", "bundle", and "bundle_lid". Multiple values may be
+                         specified separated by spaces.
+    --config-file CONFIG_FILE
+                         An optional .ini configuration file for further customization.
+Example:
+python3 pds4_create_xml_index.py <toplevel_directory> "glob_path1" "glob_path2" 
+--output_file <outputfile> --elements-file sample_elements.txt --verbose
+"""
+
 import argparse
 import configparser
 from lxml import etree
@@ -55,6 +99,14 @@ def default_value_for_nil(config, data_type, nil_value):
 
 
 def grab_elements(xpath):
+    """Extract elements from an XPath in the order they appear.
+
+    Inputs:
+        xpath    The XPath of a scraped element
+
+    Returns:
+        The tuple of elements the XPath is composed of.
+    """
     elements = ()
     parts = xpath.split('/')
 
@@ -103,7 +155,7 @@ def load_config_file(specified_config_file):
 def process_schema_location(file_path):
     """Process schema location from an XML file.
 
-    Args:
+    Inputs:
         file_path    Path to the XML file.
 
     Returns:
@@ -383,7 +435,7 @@ def main():
 
         for key in list(xml_results.keys()):
             process_tags(xml_results, key, root,
-                         namespaces, prefixes, args)
+                         namespaces, prefixes)
 
         if args.disambiguate_xpaths:
                 elements = ()
@@ -427,4 +479,3 @@ def main():
 
 if __name__ == '__main__':
     main()
-    
--- a/pds4indextools/pds4_create_xml_index.py
+++ b/pds4indextools/pds4_create_xml_index.py
@@ -1,50 +1,3 @@
-"""
-XML Bundle Scraper
-
-This script scrapes XML files within specified directories, extracts information from
-user-defined XML elements, and generates a CSV index file. The script provides options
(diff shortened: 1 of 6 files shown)


pds4indextools/pds4_create_xml_index.py
```diff
@@ -1,6 +1,5 @@
 """
 XML Bundle Scraper
-
 This script scrapes XML files within specified directories, extracts information from
```
rfrenchseti, line 3: Seems nice to have the blank line after the title

pds4indextools/pds4_create_xml_index.py
```diff
@@ -25,7 +24,8 @@
                          with quotes.
     --elements-file ELEMENTS_FILE
                          Optional text file containing elements to scrape.
-    --xpaths             Activate XPath headers in the final index file.
+    --disambiguate-xpaths
+                         Replace unique XPath headers with shortened versions.
```
rfrenchseti, line 28: All XPath headers are unique. Maybe you want to say "unique XPath leaf tags" or "unique XPath segments"?

pds4indextools/pds4_create_xml_index.py
```diff
@@ -46,14 +46,46 @@
...
 import requests
 import sys
 
 
+SplitXPath = namedtuple('SplitXPath',
+                        ['xpath', 'parent', 'child', 'prefix', 'num'])
+
+
+def convert_header_to_xpath(root, xpath_find, namespaces):
+    """Convert an XML header path to an XPath expression.
```
rfrenchseti, line 64: I have no idea what this function is doing. Perhaps some more explanatory text would help someone understand the purpose? See my renumber function (the original version) for an example of giving details.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -222,40 +185,123 @@ def process_schema_location(file_path):
...
-        del xml_results[key]
+    key_new = convert_header_to_xpath(root, key, namespaces)
+    for namespace in prefixes.keys():
+        if namespace in key_new:
+            key_new = key_new.replace(
+                '{'+namespace+'}', prefixes[namespace]+':')
+    label_results[key_new] = label_results.pop(key)
+
+
+def renumber_xpaths(xpaths):
```
rfrenchseti, line 213: Why did you take out all the comments I wrote, and the additional explanatory text in the docstring? I think they were all very useful for someone looking at this code in the future and trying to figure out what it does.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -222,40 +185,123 @@ def process_schema_location(file_path):
     return xsd_urls
 
 
-def process_tags(xml_results, key, root, namespaces, prefixes, args):
+def process_tags(label_results, key, root, namespaces, prefixes):
     """Process XML tags based on the provided options.
 
     If the --xpaths command is used, the XPath is converted into a format that
```
rfrenchseti, line 191: I think this comment is out of date

pds4indextools/pds4_create_xml_index.py
```diff
@@ -434,26 +532,50 @@ def main():
...
-        xml_results = {}
-        traverse_and_store(root, tree, xml_results, elements_to_scrape,
+        label_results = {}
+        traverse_and_store(root, tree, label_results, elements_to_scrape,
                            nillable_elements_info, config, file)
 
-        for key in list(xml_results.keys()):
-            process_tags(xml_results, key, root,
-                         namespaces, prefixes, args)
+        for key in list(label_results.keys()):
```
rfrenchseti, line 539: No reason to use list() here

pds4indextools/pds4_create_xml_index.py
```diff
@@ -434,26 +532,50 @@ def main():
...
+            process_tags(label_results, key, root,
```
rfrenchseti, line 540: Does this really need two lines?

pds4indextools/pds4_create_xml_index.py
```diff
@@ -434,26 +532,50 @@ def main():
...
+                         namespaces, prefixes)
+            
+        new_xpaths = renumber_xpaths(label_results.keys())
```
rfrenchseti, line 543: I would call this xpath_map

pds4indextools/pds4_create_xml_index.py
```diff
@@ -434,26 +532,50 @@ def main():
...
+        for key, value in new_xpaths.items():
```
rfrenchseti, line 544: Maybe call key old_xpath and value new_xpath?

pds4indextools/pds4_create_xml_index.py
```diff
@@ -434,26 +532,50 @@ def main():
...
+            label_results[value] = label_results.pop(key)
+
+        if args.disambiguate_xpaths:
+                elements = ()
```
rfrenchseti, line 548: Bad indent in this block

pds4indextools/pds4_create_xml_index.py
```diff
@@ -434,26 +532,50 @@ def main():
...
+                xpath_elements = []
+                tags = []
+                for key in list(label_results.keys()):
```
rfrenchseti, line 551: No reason to use list() here

pds4indextools/pds4_create_xml_index.py
```diff
@@ -75,42 +107,24 @@ def default_value_for_nil(config, data_type, nil_value):
     return default
 
 
-def convert_header_to_tag(path, root, namespaces):
-    """Convert an XPath expression to an XML tag.
+def grab_elements(xpath):
```
rfrenchseti, line 110: grab_elements doesn't really help me understand what this function does. Maybe it's split_into_elements or something like that?

pds4indextools/pds4_create_xml_index.py
```diff
@@ -434,26 +532,50 @@ def main():
...
+        if args.disambiguate_xpaths:
+                elements = ()
+                xpath_elements = []
+                tags = []
+                for key in list(label_results.keys()):
+                    xpath_elements.append(grab_elements(key))
+                
+                duplicates = [t for t in set(xpath_elements) if xpath_elements.count(t) > 1]
+
+                for key in list(label_results.keys()):
```
rfrenchseti, line 556: So here is one case where you actually DO need to use list, because you're modifying label_results inside the loop. If you don't use list() to create a copy of the keys first, you will end up modifying the contents of the dictionary while you're iterating over it, which will give bad results (and usually even a Python warning to let you know you did it).

pds4indextools/pds4_create_xml_index.py
```diff
@@ -434,26 +532,50 @@ def main():
...
+                    elements = grab_elements(key)
+                    tag = elements[-1]
+                    if elements not in duplicates and elements[-1] not in tags:
+                        value = tag
+                        tags.append(tag)
+                    else:
+                        value = key
+                    label_results[value] = label_results.pop(key)
```
rfrenchseti, line 564: Doesn't this technique change the order of the columns? Any tag that is unique will end up being at the very end of the list, right?

rfrenchseti (changes requested):
I think the sense of "--disambiguate-xpaths" is backwards. The normal behavior is that we disambiguate them by giving the full xpath. What we want to indicate is that we're going to NOT disambiguate them if they aren't ambiguous in the first place. Perhaps something like "--simplify-xpaths".

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: So, it puts the new result at the end of the dictionary. However, since it does this for ALL entries, it keeps putting the next new entry at the end, until all the entries are replaced. It's like a conveyer belt:

a-b-c-d 

b-c-d-a*

c-d-a*-b*

d-a*-b*-c*

a*-b*-c*-d*

rfrenchseti: I saw this above, but for some reason while skimming I thought this code was only executed for some of the labels, resulting in them being out of order. But I see it is done for all labels so it works as expected.

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: Okay I actually came back to this because I didn't test it correctly. `list()` is actually needed here, since otherwise the dictionary is changed in-place. I get this error if the `list` is removed:

```
  File "/Users/emiliesimpson/Github/rms-pds4indextools/pds4indextools/pds4_create_xml_index.py", line 630, in <module>
    main()
  File "/Users/emiliesimpson/Github/rms-pds4indextools/pds4indextools/pds4_create_xml_index.py", line 579, in main
    for key in label_results.keys():
RuntimeError: dictionary changed size during iteration
```

pds4indextools/pds4_create_xml_index.py
```diff
@@ -434,26 +572,49 @@ def main():
...
-        xml_results = {}
-        traverse_and_store(root, tree, xml_results, elements_to_scrape,
+        label_results = {}
+        traverse_and_store(root, tree, label_results, elements_to_scrape,
                            nillable_elements_info, config, file)
 
-        for key in list(xml_results.keys()):
-            process_tags(xml_results, key, root,
-                         namespaces, prefixes, args)
+        for key in label_results.keys():
```
rfrenchseti, line 579: It turns out you need the list() here too

Issue #8, opened 2024-03-27
`--dump-available-xpaths`: a new command that creates a file with all of the possible columns available in a file or set of files


esimpsons3ti:
If a user needs to generate an index file from numerous labels, it may become lengthy. The new command line argument, --dump-available-xpaths, outputs a file with all elements gathered by the tool to a specified location. This feature enables users to review scraped label information and refine their query accordingly.

PR #10, opened 2024-04-09
Adding unit tests, golden files, and supplemental labels

### LINKED ISSUE #5: Need "golden copy" labels
For the indexing tools unit tests, "golden copy" versions of labels and index files are needed to compare current results against. 

The "dummy" labels will be representative of PDS4 labels, but are not replicas of existing labels. They will vary in length and content.

### LINKED ISSUE #6: Need "golden copy" index files
These "golden copies" of index files will serve as benchmarks against which the results of the unit tests will be compared. They will vary in terms of content, length, and context. Each "golden copy" will represent several aspects of the index tools.

### LINKED ISSUE #7: Write unit tests for the index tools
The indexing tools need unit tests to ensure that the generated results are accurate and consistent. The unit tests will cover:
- multiple combinations of labels
- all command-line options
- several success and failure instances

esimpsons3ti:
This pull request will introduce the unit tests for the indexing tools.

Fixes #7

Fixes #6 

Fixes #5


esimpsons3ti:
This is now in a state to be actually looked at. The changes also now include new functionality in the current program to do the following:
- Now reads elements file as glob statements
- square brackets are replaced by angled brackets
- Comments within label are now removed automatically
- And more!

Let me know if you have any questions.

esimpsons3ti:
I have added the new command line argument function, now named `--dump-available-elements`.

Fixes #8

pds4indextools/pds4_create_xml_index.py
```diff
@@ -75,42 +114,46 @@ def default_value_for_nil(config, data_type, nil_value):
...
-    xpath_final = ''
-    portion = ''
-    for sec in sections[1:]:
-        portion = portion + '/' + sec
-        tag = str(root.xpath(portion, namespaces=namespaces)[0].tag)
-        xpath_final = xpath_final + '/' + tag
+    filtered_dict = {}
+
+    if glob_patterns is not None:
+        if not all(pat.startswith('!') for pat in glob_patterns):
```
rfrenchseti, line 130: This doesn't work. The idea is that the glob patterns are executed *in order*. For example:
```
a*
!ab*
abc*
```
would match anything starting with a, unless it started with ab, but starting with abc is still OK. If you break the patterns into separate positive/negative lists and then do all the positives and all the negatives, you don't get this behavior. Just do the patterns in order. The code can be much simpler.

Also, I think you're treating the case of:
```
!a
```
as meaning:
```
*
!a
```
(hence the use of `all()` and the separate if/else) but that's not what the user said. If the user just says:
```
!a
```
then they should get no elements.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -146,70 +189,22 @@ def load_config_file(specified_config_file):
...
     Returns:
         List of XSD URLs extracted from the schema location.
     """
     # Load and parse the XML file
-    tree = etree.parse(file_path)
-    root = tree.getroot()
+    try:
+        tree = etree.parse(file_path)
+        root = tree.getroot()
+    except OSError:
```
rfrenchseti, line 206: Do both of these statements potentially trigger an `OSError`? If not, the other should be outside the try/except block.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -222,40 +217,190 @@ def process_schema_location(file_path):
...
 
-    If the --xpaths command is used, the XPath is converted into a format that
-    contains the names and namespaces of all the parent elements of that element.
-    If the --xpaths command is not used, the XPath is converted into the
-    associated element tag of that element, and given its associated namespace. These
-    values then replace their old versions in the xml_results dictionary.
+    Processes XPath headers by converting parts of the XPath into element tags,
+    replacing namespaces with prefixes, and updating the label_results dictionary.
+    If a duplicate XPath is encountered, it appends an underscore and a number
+    to make the XPath unique.
```
rfrenchseti, line 226: Oh maybe this answers my question below about where the `_1` comes from in the SPICE kernel list. I don't understand how this can happen. Why is there a case where you get `_1` instead of `<1>`?

pds4indextools/pds4_create_xml_index.py
```diff
@@ -222,40 +217,190 @@ def process_schema_location(file_path):
...
+        i = 1
+        while True:
+            new_key = f"{key_new}_{i}"
+            if new_key not in label_results:
+                key_new = new_key
+                break
+            i += 1
+    label_results[key_new] = label_results.pop(key)
+
+
```
rfrenchseti, line 251: Too many blank lines

pds4indextools/pds4_create_xml_index.py
```diff
@@ -222,40 +217,190 @@ def process_schema_location(file_path):
...
-        del xml_results[key]
-    else:
-        key_new = convert_header_to_tag(key, root, namespaces)
-        for namespace in prefixes.keys():
-            if namespace in key_new:
-                key_new = key_new.replace(
-                    '{'+namespace+'}', prefixes[namespace]+':')
-        xml_results[key_new] = xml_results[key]
-        del xml_results[key]
+    elements = ()
```
rfrenchseti, line 395: This is a real nit, but since tuples are immutable they are not meant to be appended to. If you append an object to a tuple with `+=`, it creates a whole new tuple from scratch, copying over all the old data. However, lists are mutable and are meant to be extended. So `x.append(y)` is an efficient operation, while `x += (y,)` (where x is a tuple) is less efficient.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -222,40 +217,190 @@ def process_schema_location(file_path):
...
+    parts = xpath.split('/')
+
+    for part in parts:
+        if '<' in part:
+            part = part.split('<')
+            elements += (part[0],)
```
rfrenchseti, line 401: Do you really only want to include elements that have `<n>` at the end?

pds4indextools/pds4_create_xml_index.py
```diff
@@ -434,35 +637,95 @@ def main():
...
+        elements = []
+        for label in all_results:
+            for values in label.values():
+                for x in values.keys():
+                    elements.append(x)
+        with open(output_path, 'w') as file:
+            for item in elements:
+                file.write("%s\n" % item) 
+    else:
+
```
rfrenchseti, line 726: Remove blank line

pds4indextools/pds4_create_xml_index.py
```diff
@@ -75,42 +114,46 @@ def default_value_for_nil(config, data_type, nil_value):
...
+        if not all(pat.startswith('!') for pat in glob_patterns):
+            negated = []
+            positive = []
+            for pat in glob_patterns:
+                if pat.startswith('!'):
+                    pat = pat.replace('!', '')
+                    negated.append(pat)
+                else:
+                    positive.append(pat)
+            for key, value in input_dict.items():
```
rfrenchseti, line 139: One thing that might be useful here is allowing the "verbose" option to give details about what is happening here. For example, you could print:
```
Adding elements due to: **/*a*
  /pds:/whatever
  /pds:/whatever
Deleting elements due to: !**/*ab*
  /whatever
```

rfrenchseti (changes requested):
- I haven't tried to run the code, so these comments are just from visual inspection.
- I know it's a pain, but give some consideration to what I suggested earlier about creating the golden files by hand *before* running the tests. Right now I think the golden files are being generated by the code, and there are cases where they are wrong. If you had written them by hand first, you would have written the correct answer and then found a bug when running the test. It takes a lot longer to do than just auto-generating them and looking at them to see if you think they're correct, but you will catch more bugs this way.
- All files related to testing should go under the single `test_files` directory. There shouldn't be a separate top-level `samples` directory. Then you can create sub-directories under there for different purposes. For example, `test_files/sample_input` and `test_files/expected`. Then you don't need to add `_success` to the end of your filenames, although it doesn't hurt to leave it there.
- Your input files seem to be split between two locations, e.g. `test_files/tester_label_1.xml` and `samples/sample_elements.txt`. You could put all these under `test_files/sample_input` or divide them further like `test_files/sample_input/element_files` and `test_files/sample_input/labels`.
- The clean headers option needs to do something with `<` and `>` since those are also illegal identifiers in strings. I think it would be easiest just to replace `<1>` with `_1`.

test_files/elements_file_success_2.csv
```diff
@@ -0,0 +1,2 @@
+pds:Product_Observational/pds:Observation_Area<1>/pds:Discipline_Area<1>/geom:Geometry<1>/geom:SPICE_Kernel_Files<1>/geom:SPICE_Kernel_Identification<1>/geom:spice_kernel_file_name<1>,pds:Product_Observational/pds:Observation_Area<1>/pds:Discipline_Area<1>/geom:Geometry<1>/geom:SPICE_Kernel_Files<1>/geom:SPICE_Kernel_Identification<1>/geom:spice_kernel_file_name_1<1>,pds:Product_Observational/pds:Observation_Area<1>/pds:Discipline_Area<1>/geom:Geometry<1>/geom:SPICE_Kernel_Files<1>/geom:SPICE_Kernel_Identification<1>/geom:spice_kernel_file_name_2<1>,pds:Product_Observational/pds:Observation_Area<1>/pds:Discipline_Area<1>/geom:Geometry<1>/geom:SPICE_Kernel_Files<1>/geom:SPICE_Kernel_Identification<1>/geom:spice_kernel_file_name_3<1>,pds:Product_Observational/pds:Observation_Area<1>/pds:Discipline_Area<1>/geom:Geometry<1>/geom:SPICE_Kernel_Files<1>/geom:SPICE_Kernel_Identification<1>/geom:spice_kernel_file_name_4<1>
```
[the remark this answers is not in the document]
esimpsons3ti: This is done on behalf of the code, and is not present within the label file. It is a way to make each duplicate XPath artificially unique so that it can catch elements with otherwise identical XPaths. Since we already didn't care about the XPath headers following a parseable syntax, I didn't consider it an issue in the final product. 

If you prefer the trailing underscores and numbers to be removed in the final product (i.e. when all the desired contents are already in the index file) I can do so.

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: So, the issue lies not in the predicates (turns out that's what the `[1]` are called in these), but in the namespaces that are used. The `geom` namespace does not use predicates. And since these numbers don't exist until they are artificially added, nothing stops them from overwriting each other. The added underscore and number are just to keep everything in order until the XPaths get the predicates and go through renumbering.

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: At this point in the code, all elements within every XPath header should have `<n>` at the end.

tests/test_pds4_create_xml_index.py
```diff
@@ -0,0 +1,460 @@
...
+            ),
+
+            (str(test_files_dir / 'simplify_xpaths_success_3.csv'),
+             'simplify_xpaths_3.csv',
+             [
+                str(test_files_dir),
+                'tester_label_2.xml',
+                '--simplify-xpaths',
+                '--elements-file',
+                str(root_dir / 'samples/elements_xpath_simplify_2.txt'),
```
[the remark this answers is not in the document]
esimpsons3ti: I have fixed the numbering so that everything is consistent

tests/test_pds4_create_xml_index.py
```diff
@@ -0,0 +1,460 @@
...
+             'simplify_xpaths_1.csv',
+             [
+                str(test_files_dir),
+                'tester_label_1.xml',
+                '--simplify-xpaths',
+                '--output-file',     
+             ]
+            ),
+
+            (str(test_files_dir / 'simplify_xpaths_success_2.csv'),
```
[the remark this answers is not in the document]
esimpsons3ti: I agree that using `--dump-available elements` in certain unit tests is a better solution. This will be applied to the unit tests that can benefit from this feature (especially `--simplify-xpaths`)

test_files/simplify_xpaths_success_1.csv
```diff
@@ -0,0 +1,2 @@
+pds:logical_identifier,pds:version_id,pds:title,pds:information_model_version,pds:name,pds:Product_Observational/pds:Observing_System<1>/pds:Observing_System_Component<1>/pds:name<1>,pds:type,pds:lid_reference,pds:reference_type
```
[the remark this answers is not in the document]
esimpsons3ti: Thank you for catching this! This functionality has been fixed.

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: This has been fixed.

tests/test_pds4_create_xml_index.py
```diff
@@ -0,0 +1,460 @@
...
+        'golden_file,new_file,cmd_line',
+        [
+            # simple test cases
+            (str(test_files_dir / 'elements_file_success.csv'),
+             'elements_file.csv',
+             [
+                str(test_files_dir),
+                'tester_label_1.xml',
+                '--elements-file',
+                str(root_dir / 'samples/sample_elements.txt'),
```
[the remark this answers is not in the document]
esimpsons3ti: The no-glob test will be considered under failures, since this results in an entirely blank index file.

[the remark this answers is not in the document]
rfrenchseti: I don't mean that there are no entries in the file, I mean that there are no glob *characters*, like `*` or `**`. Just a list of plain XPaths with no wildcards.

But I wouldn't consider a blank output file a "failure". The user asked for a blank file, they got a blank file - success! It's only a failure if the program doesn't do what it's expected to do.

tests/test_pds4_create_xml_index.py
```diff
@@ -0,0 +1,460 @@
...
+                    
+def test_elements_file(golden_file, new_file, cmd_line):
+    # Create a temporary directory in the same location as the test_files directory
+    with tempfile.TemporaryDirectory(dir=test_files_dir.parent) as temp_dir:
+        temp_dir_path = Path(temp_dir)
+        
+        # THE PATH TO THE NEW FILE
+        path_to_file = temp_dir_path / new_file
+        # Call main() function with the simulated command line arguments
+        cmd_line.append(str(path_to_file))
```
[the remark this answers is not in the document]
esimpsons3ti: Mostly this is from the original example Dave provided. However, now that we're using `--dump-available-elements` in some of the unit tests, there can be either a `csv` or a `txt` temp file. We could:
- keep it as is
- Change all the successful files to one file type.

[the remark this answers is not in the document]
rfrenchseti: I'll leave that decision up to you. Having a single output filename (probably `.txt`) seems like it would make the code 
simpler, but it's not a big deal either way.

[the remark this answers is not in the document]
esimpsons3ti: Keeping it as is, then.

PR #15, opened 2024-06-17
PDS4 Indexing Tools Improvements and Additions

### LINKED ISSUE #3: Write a PDS4 label for the output table
From rms-data-projects created by [matthewtiscareno](https://github.com/matthewtiscareno): SETI/rms-data-projects#71  

This tool produces an output table that is PDS4-compliant.  In order to fully incorporate an output table into the PDS4 holdings, a label is required. 

Design a template label for output tables, and then add an option for the tool to automatically generate such a label.

### LINKED ISSUE #5: Need "golden copy" labels

### LINKED ISSUE #6: Need "golden copy" index files

### LINKED ISSUE #8: `--dump-available-xpaths`: a new command that creates a file with all of the possible columns available in a file or set of files

esimpsons3ti:
1. The indexing tool (now `pds4_create_xml_index.py`) now has six new argparse commands:
```
    --output-headers-file
    --limit-xpaths-file (previously --elements-file)
    --simplify-xpaths (previously --disambiguate-xpaths)
    --fixed-width
    --generate-label
    --label-user-input
```

2. Major changes have been implemented into the code to make the result more human-readable and to fix several major bugs. Some of these issues include missing indexed data, inaccurate XPath headers, and Windows incompatibility.

3. Added unit tests and associated supplemental files, which as of 7/5 all pass:
    - `tests/` contains the unit tests for the indexing tool
    - `test_files/expected` contains golden copies for unit test comparison
    - `test_files/samples` contains supplemental files for the unit tests to use
    - `test_files/labels` contains fake label files for the unit tests to use

5. Docstrings and internal documentation has been fully incorporated into the indexing tool, and has also been reformatted to follow Google style.

There is also the implementation of the label generation feature, which includes:
    - label generation functions
    - template label file
    - default input file (called `default_values.yaml`)

- Fixes #3
- Fixes #5
- Fixes #6
- Fixes #8


pds4indextools/pds4_create_xml_index.py
```diff
@@ -75,56 +154,80 @@ def default_value_for_nil(config, data_type, nil_value):
...
-        path          XPath expression.
-        root          The root element of the XML document.
-        namespaces    Dictionary of XML namespace mappings.
+    Args:
+        tree (ElementTree.Element): The XML tree.
 
     Returns:
-        Converted XML tag.
+        str or None: The text content of the logical_identifier element,
+                     or None if not found.
```
rfrenchseti, line 166: This is fine, but just FYI the normal indentation for multi-line Args or Returns is to wrap the text with a 4-space indent, like:

```
        str or None: The text content of the logical_identifier element,
            or None if not found.
```

That way you don't have to squeeze everything into a narrow column on the right.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -222,52 +278,209 @@ def process_schema_location(file_path):
...
+        prefixes (dict): A dictionary containing XML namespace prefixes.
+    """
+    key_new = convert_header_to_xpath(root, key, namespaces)
+    for namespace in prefixes.keys():
+        if namespace in key_new:
+            key_new = key_new.replace(
+                '{'+namespace+'}', prefixes[namespace]+':')
+    if key_new in label_results:
+        # If the XPath already exists, append an underscore and a number
+        i = 1
```
rfrenchseti, line 304: How about `suffix_num`? As discussed below we generally want to avoid single-character variable names except inside list comprehensions.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -222,52 +278,209 @@ def process_schema_location(file_path):
...
+        while True:
+            new_key = f"{key_new}_{i}"
```
rfrenchseti, line 306: Having both `new_key` and `key_new` is confusing. How about `trial_key`?

pds4indextools/pds4_create_xml_index.py
```diff
@@ -222,52 +278,209 @@ def process_schema_location(file_path):
...
+            if new_key not in label_results:
+                key_new = new_key
+                break
+            i += 1
```
rfrenchseti, line 310: This code is fine, but if you want something a little more Pythonic where you don't have to keep track of the index separately, you can use `itertools`:
```
for suffix_num in itertools.count(start=1, step=1):
    <body>
```

pds4indextools/pds4_create_xml_index.py
```diff
@@ -222,52 +278,209 @@ def process_schema_location(file_path):
...
-    If the --xpaths command is used, the XPath is converted into a format that
-    contains the names and namespaces of all the parent elements of that element.
-    If the --xpaths command is not used, the XPath is converted into the
-    associated element tag of that element, and given its associated namespace. These
-    values then replace their old versions in the xml_results dictionary.
+    Args:
+        xpaths (list): The list of XPaths or XPath fragments.
 
+    Returns:
+        dict: A dictionary containing a mapping from the original XPaths to the renumbered XPaths.
```
rfrenchseti, line 368: Wrap to 90 chars

pds4indextools/pds4_create_xml_index.py
```diff
@@ -222,52 +278,209 @@ def process_schema_location(file_path):
...
+    # Group split_xpaths by prefix
+    for prefix, prefix_group in groupby(split_xpaths, lambda x: x.prefix):
+        prefix_group_list = list(prefix_group)
+
+        # The parents in the resulting group may have unique IDs.
+        # We collect those IDs and create a mapping from the original numbers
+        # to a new set of suffixes of the form "[<n>]" where <n> is sequentially
+        # increasing starting at 1. We also add a special entry for the empty
+        # suffix when there is no number.
+        unique_nums = sorted(list(set(x.num for x in prefix_group_list
```
rfrenchseti, line 416: Optimizing my own code, we can use a set comprehension:
```
        unique_nums = sorted({x.num for x in prefix_group_list if x.num is not None})
```

pds4indextools/pds4_create_xml_index.py
```diff
@@ -222,52 +278,209 @@ def process_schema_location(file_path):
...
+
+    for part in parts:
+        if '<' in part:
+            part = part.split('<')
+            elements.append(part[0])
+
+    return elements
 
 
 def store_element_text(element, tree, results_dict, nillable_elements_info, config, label):
```
rfrenchseti, line 473: Line too long

pds4indextools/pds4_create_xml_index.py
```diff
@@ -222,52 +278,209 @@ def process_schema_location(file_path):
...
-        config                    The configuration data.
-        label                     The name of the label file.
+    """
+    Store text content of an XML element in a results dictionary.
+
+    Args:
+        element (Element): The XML element.
+        tree (ElementTree): The XML tree.
+        results_dict (dict): Dictionary to store results.
+        nillable_elements_info (dict): A dictionary containing nillable element information.
```
rfrenchseti, line 481: Line too long

pds4indextools/pds4_create_xml_index.py
```diff
@@ -222,52 +278,209 @@ def process_schema_location(file_path):
...
+        config (dict): The configuration data.
+        label (str): The name of the label file.
```
rfrenchseti, line 483: Maybe call this `label_filename`? Same in next function.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -222,52 +278,209 @@ def process_schema_location(file_path):
...
-        nillable_elements_info    A dictionary containing nillable element information.
-        config                    The configuration data.
-        label                     The name of the label file.
+    """
+    Store text content of an XML element in a results dictionary.
+
+    Args:
+        element (Element): The XML element.
+        tree (ElementTree): The XML tree.
+        results_dict (dict): Dictionary to store results.
```
rfrenchseti, line 480: Dictionary in which to store results

pds4indextools/pds4_create_xml_index.py
```diff
@@ -293,60 +506,149 @@ def store_element_text(element, tree, results_dict, nillable_elements_info, conf
...
+    store_element_text(element, tree, results_dict,
+                       nillable_elements_info, config, label)
     for child in element:
-        traverse_and_store(child, tree, results_dict, elements_to_scrape,
+        traverse_and_store(child, tree, results_dict,
                            nillable_elements_info, config, label)
 
 
+@functools.lru_cache(maxsize=None)
+def download_xsd_file(xsd_file):
```
rfrenchseti, line 532: Is `xsd_file` really a URL? In which case maybe call it `xsd_file_url`.

rfrenchseti: Also add docstring

pds4indextools/pds4_create_xml_index.py
```diff
@@ -293,60 +506,149 @@ def store_element_text(element, tree, results_dict, nillable_elements_info, conf
...
+        sys.exit(1)
+
+
+def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
+    """
+    Store all nillable elements and their data types in a dictionary.
+
+    Args:
+        xsd_file (str): An XML Schema Definition file.
+        nillable_elements_info (dict): A dictionary containing nillable element information.
```
rfrenchseti, line 546: Line too long

pds4indextools/pds4_create_xml_index.py
```diff
@@ -293,60 +506,149 @@ def store_element_text(element, tree, results_dict, nillable_elements_info, conf
...
-    """Write results from a list of dictionaries to a CSV file.
+    """
+    Write results from a list of dictionaries to a CSV file.
 
-    Inputs:
-        results_list          List of dictionaries containing results.
-        output_csv_path       The output directory and filename.
+    Args:
+        results_list (list): List of dictionaries containing results.
+        output_csv_path (str): The output directory and filename.
```
rfrenchseti, line 599: Missing `args` parameter

pds4indextools/pds4_create_xml_index.py
```diff
@@ -293,60 +506,149 @@ def store_element_text(element, tree, results_dict, nillable_elements_info, conf
...
         df.rename(columns=lambda x: x.replace(
-            ':', '_').replace('/', '__'), inplace=True)
+            ':', '_').replace('/', '__').replace('<', '_').replace('>', ''), inplace=True)
+        
+    if args.fixed_width:
+        padded_df = pad_column_values_and_headers(df)
+        padded_df.to_csv(output_csv_path, index=False, na_rep='NaN')
 
-    df.to_csv(output_csv_path, index=False, na_rep='NaN')
+    else:
```
rfrenchseti, line 643: Extra blank line

rfrenchseti (changes requested):
- These are comments just looking at the code. I'll run it next.
- Because GitHub only lets me comment on code near lines that have changed, if I want to comment on some other part of the code that hasn't changed recently, I have to put the comment on some code nearby. So check which code I'm talking about in the comments.
- You can set your auto-formatter to use 90-character lines. There are a lot of line breaks that don't need to be there.
- Take a look at the `SETI/rms-solar` repo and copy over the top-level `.flake8` file and remove the three extend-ignore exceptions. Update the `.github/workflows/run-tests.yml` file to have the new "flake8" job (see `rms-solar`). Add `flake8` to `requirements.txt` and install it in your venv. Then run from the command line `flake8 pds4indextools tests` to see how compliant your code is. It looks pretty good to me - there are only a couple of minor things to change to make it 100% compliant. From this point forward when you check in new code it will automatically run flake8 and fail the tests if the code isn't compliant.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -359,9 +661,9 @@ def main():
...
 
-    parser.add_argument('--xpaths', action='store_true',
-                        help='If specified, use full XPaths in the column '
-                             'headers. If not specified, use only elements tags.')
+    parser.add_argument('--simplify-xpaths', action='store_true',
+                        help='If specified, uses tags of unique XPaths. Any values with '
+                             'duplicate values will still use their full XPath.')
 
     parser.add_argument('--output-file', type=str,
                         help='The output filepath ending with your chosen filename for '
```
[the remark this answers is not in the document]
rfrenchseti: Maybe my comment above isn't great:

Replace the ":" and "/" in the column headers with (legal friendly) characters can be made more specific: Replace ":" and "/" in column headers with underscores

since some are replaced with a single underscore, some with two, and some deleted entirely. Perhaps something like "Rename column headers such that they only contain characters permissible in variable names."

rfrenchseti:
I think maybe `--output-file` should be a required argument. Right now it defaults to putting the resultant CSV file in the root directory of the bundle/collection, and calling it `index_file.csv`. This is almost certainly the wrong thing to do. First of all, people probably don't want to modify their archives unless they know the result is good. Second they are unlikely to want to just call it `index_file.csv`. And if you don't know what the default is, the file is actually really hard to find. So maybe we should just force people to specify the filename/path.

rfrenchseti:
Here's another thought. Right now `--output-file` is generic - it will produce a CSV file or a TXT file full of elements. Maybe it would make more sense to have different command line options for different outputs, like `--output-csv-file` and `--output-xpaths-file` and then get rid of `--dump-available-xpaths` as an option. Then you could actually output both a CSV file and an XPath file at the same time, which could be handy. If you don't specify any output file, print a message and exit.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -428,41 +742,132 @@ def main():
...
+
+        correct_duplicates(label_results)
 
-        lid = xml_results.get('pds:logical_identifier', 'Missing_LID')
+        verboseprint(
+            'Now filtering label results according to given element file.')
+        label_results = filter_dict_by_glob_patterns(
+            label_results, elements_to_scrape, verboseprint)
+
+        if args.simplify_xpaths:
```
rfrenchseti, line 790: MAJOR BUG HERE. This can't be done on a per-label basis. It has to be done for the entire collection of labels once at the end. For example, consider the elements file:
```
*Science_Facets*
```
and the command line:
```
python pds4_create_xml_index.py /mnt/rms-holdings/pds4-holdings/bundles/uranus_occs_earthbased/uranus_occ_u0_kao_91cm "**/rings/*.xml" "**/global/*.xml"  --output-file x.txt --dump-available-xpaths  --elements-file elements.txt
```
The resulting xpath file is:
```
pds:Product_Observational/pds:Observation_Area<1>/pds:Primary_Result_Summary<1>/pds:Science_Facets<1>/pds:wavelength_range<1>
pds:Product_Observational/pds:Observation_Area<1>/pds:Primary_Result_Summary<1>/pds:Science_Facets<1>/pds:discipline_name<1>
pds:Product_Observational/pds:Observation_Area<1>/pds:Primary_Result_Summary<1>/pds:Science_Facets<1>/pds:facet1<1>
pds:Product_Observational/pds:Observation_Area<1>/pds:Primary_Result_Summary<1>/pds:Science_Facets<2>/pds:wavelength_range<1>
pds:Product_Observational/pds:Observation_Area<1>/pds:Primary_Result_Summary<1>/pds:Science_Facets<2>/pds:discipline_name<1>
pds:Product_Observational/pds:Observation_Area<1>/pds:Primary_Result_Summary<1>/pds:Science_Facets<2>/pds:facet1<1>
```
but now look what happens when I add `--simply-xpaths` to it:
```
pds:wavelength_range
pds:discipline_name
pds:facet1
pds:Product_Observational/pds:Observation_Area<1>/pds:Primary_Result_Summary<1>/pds:Science_Facets<1>/pds:wavelength_range<1>
pds:Product_Observational/pds:Observation_Area<1>/pds:Primary_Result_Summary<1>/pds:Science_Facets<1>/pds:discipline_name<1>
pds:Product_Observational/pds:Observation_Area<1>/pds:Primary_Result_Summary<1>/pds:Science_Facets<1>/pds:facet1<1>
pds:Product_Observational/pds:Observation_Area<1>/pds:Primary_Result_Summary<1>/pds:Science_Facets<2>/pds:wavelength_range<1>
pds:Product_Observational/pds:Observation_Area<1>/pds:Primary_Result_Summary<1>/pds:Science_Facets<2>/pds:discipline_name<1>
pds:Product_Observational/pds:Observation_Area<1>/pds:Primary_Result_Summary<1>/pds:Science_Facets<2>/pds:facet1<1>
```
We have some xpaths simplified and some not simplified.

You should check other things that you're doing on a per-label basis to make sure they don't need to be done after all the labels are read as well.

rfrenchseti: This is something that should have been caught by the test suite, so make sure you add new tests to cover this kind of thing.

rfrenchseti:
The test coverage is only 85%, and if you look at the coverage HTML report you'll see a lot of code that really needs to be tested. It also shows some code that might be obsolete and never used anymore, in which case you should delete it. I think this program should be at 100% test coverage. It won't take that much effort to get there.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -359,9 +661,9 @@ def main():
                              'specified, all elements found in the XML files are '
```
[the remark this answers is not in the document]
esimpsons3ti: I have replaced it with 'label files', since it should be compatible with .xml or .lblx.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -1,21 +1,27 @@
...
+the XPaths available to the user. The script provides options for customizing the 
+extraction process, such as:
+    - Specifying desired content by either limiting the results or requesting additional
+      file information.
+    - Sorting the resulting file by a user-specified value.
+    - Allowing for a user-made configuration file (.ini) for further custom content.
 
 Usage:
-    python xml_bundle_scraper.py <directorypath> <pattern>
+    python pds4_create_xml_index.py <directorypath> <pattern>
```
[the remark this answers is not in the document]
esimpsons3ti: Considering the term 'eventually' here, I'm going to keep this unchecked until the very end of development.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -428,41 +742,132 @@ def main():
...
+    if args.dump_available_xpaths:
+        verboseprint(f'XPaths file generated at {output_path}')
+        xpaths = []
+        for label in all_results:
+            for values in label.values():
+                for x in values.keys():
+                    if x not in xpaths:
+                        xpaths.append(x)
+
+        for x in xpaths:
```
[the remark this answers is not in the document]
esimpsons3ti: This code is actually deprecated. The functionality it was intended for is actually covered by the `correct_duplicates()` function. I just forgot to remove this little bit of code. It has been removed.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -25,45 +31,118 @@
...
+    Returns:
+        str: Converted XPath expression.
+    """
+    sections = xpath_find.split('/')
+    xpath_final = ''
+    portion = ''
+    for sec in sections[1:]:
+        portion = portion + '/' + sec
+        tag = str(root.xpath(portion, namespaces=namespaces)[0].tag)
+        if '*' in sec:
```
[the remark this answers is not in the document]
esimpsons3ti: If the unprocessed XPath headers always followed the same syntax, then this would apply. However, this isn't the case. Below are a collection of XPath headers when they are first scraped with `lxml`:

```
/*/*[2]/*[6]/*[2]
/*/*[2]/*[6]/*[3]/*[1]
/*/*[2]/*[6]/*[3]/*[2]
/*/*[2]/*[7]/cassini:Cassini/cassini:ISS_Specific_Attributes/cassini:mission_phase_name
/*/*[2]/*[7]/cassini:Cassini/cassini:ISS_Specific_Attributes/cassini:spacecraft_clock_count_partition
/*/*[2]/*[7]/cassini:Cassini/cassini:ISS_Specific_Attributes/cassini:spacecraft_clock_start_count
```

depending on the namespace, `lxml` will give some XPaths different formats. The ones that are just asterisks and predicates are usually the default namespace (`pds`), while the half-and-half syntax is for any other namespace. 

In short, `sec.startswith('*') ` can apply here, but `sec.startswith(':')` doesn't. It needs to remain `if ':' in sec:`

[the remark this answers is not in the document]
rfrenchseti: Got it. It was mainly the first one I was concerned about. It seemed like a potential bug that you look for a * anywhere, but then delete the first character even if it's potentially not a *. Using `startswith` guarantees that you're always deleting a *.

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: Oh! That actually helps quite a bit.

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: Thank you for catching this. The new code works with your example, and all tests now pass.

matthewtiscareno:
It seems that, if there is an attribute that appears in some of the indicated product labels but not others, then the column will be created and the value of "NaN" will be entered in that column for those product labels that do not include the attribute.  Is this intended?

rfrenchseti:
> It seems that, if there is an attribute that appears in some of the indicated product labels but not others, then the column will be created and the value of "NaN" will be entered in that column for those product labels that do not include the attribute. Is this intended?

As a side note, it's a very bad idea to create index files from labels that have dramatically different layouts. The resulting index will just have a lot of empty columns and won't make much sense. This is why, for example, we create three different index files for `uranus_occs_earthbased` - even though rings, global, and atmosphere have a lot in common, they are different enough that having a single index file is just confusing. I'm mentioning this here mainly so that @esimpsons3ti will remember to put this as a big important note in the documentation she will write.

That said, there are invariably small changes between labels in the same data set. For example, there might be a different number of keywords, or SPICE kernels, or auxiliary products. So we need to be able to handle those cases.

The question is: What do we put in the CSV file when the data is "missing"? The current code puts in `NaN` for both string and numeric fields, which I agree is not great. But what should we put in? I would argue that maybe the fields should just be empty. Alternatively we could make it used-configurable.

esimpsons3ti:
A few notes about this:

- This behavior was due to an element within the batch of tested label files being nilled when they weren't nillable (in this case, `Complex_FOV`). Previously, I didn't have code that would go in and look for the data type of the erroneously-nilled element. It would just print a warning and move on.
- Now that I do have this code, I have integrated it into `store_element_text`. Now, if a non-nillable element is nilled, it will still print the warning, but there will no longer be NaNs placed where the default value is.

rfrenchseti:
I don't think this is the only time this can happen. I produced NaNs myself using the Uranus occ bundles.

matthewtiscareno:
The Cassini VIMS bundle contains a number of "composite structures" that are broken up into individual data products.  The data products that are part of a composite structure have an entire additional class full of attributes that describes how they relate to the rest of the structure.  For example, in [this directory](https://pds-rings.seti.org/pds4/bundles/cassini_vims_cruise/data_raw/13089xxxxx), the data products ending with _xxx have that additional class, while those that don't don't. 

It is quite reasonable to want an index file that includes all of these products together, which means that the case arises of columns that are present in some labels but absent from others. 

Do we need to inquire into what is the data type of each column in order to define the special constants for when they are missing?

esimpsons3ti:
flake8 has been failing the code on line length, even when my flake8.txt file is set to the 90 character limit. But I'll wait on your feedback for now.

pds4indextools/default_values.yaml
```diff
@@ -0,0 +1,21 @@
+{'version_id': '1.1',
```
rfrenchseti, line 1: While JSON is generally parsable by YAML readers, the main advantage of YAML is it's inline format, which is easier to write and read by humans. You might consider changing this to:
```
- version_id: '1.1'
- title: 'Tested Index File'
```

Also there are a lot of blank lines at the bottom of the file.

pds4indextools/template_pds.xml
```diff
@@ -0,0 +1,178 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rfrenchseti, line 1: Maybe call this file `index_label_template_pds.xml` to make it clear it's for the index labels?

pds4indextools/template_pds.xml
```diff
@@ -0,0 +1,178 @@
...
+<?xml-model href="https://pds.nasa.gov/pds4/pds/v1/PDS4_PDS_1L00.sch"
+    schematypens="http://purl.oclc.org/dsdl/schematron"?>
+
+$IF(Product_Ancillary)   
+<Product_Ancillary xmlns="http://pds.nasa.gov/pds4/pds/v1"
+ xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+ xsi:schemaLocation="http://pds.nasa.gov/pds4/pds/v1 https://pds.nasa.gov/pds4/pds/v1/PDS4_PDS_1L00.xsd">
+$END_IF
+$IF(Product_Metadata_Supplemental)
```
rfrenchseti, line 10: I'm guessing that either `Product_Ancillary` is True or `Product_Metadata_Supplemental` is True, but not neither and not both. You could simplify this by simply testing for one of them and using `$ELSE`. Or maybe add tests for both and neither just to catch the case where the template is being fed bad metadata?

pds4indextools/template_pds.xml
```diff
@@ -0,0 +1,178 @@
...
+ xsi:schemaLocation="http://pds.nasa.gov/pds4/pds/v1 https://pds.nasa.gov/pds4/pds/v1/PDS4_PDS_1L00.xsd">
+$END_IF
+$IF(Product_Metadata_Supplemental)   
+<Product_Metadata_Supplemental xmlns="http://pds.nasa.gov/pds4/pds/v1"
+ xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+ xsi:schemaLocation="http://pds.nasa.gov/pds4/pds/v1 https://pds.nasa.gov/pds4/pds/v1/PDS4_PDS_1L00.xsd">
+$END_IF
+    <Identification_Area>
+        <logical_identifier>$logical_identifier$</logical_identifier>                          
+        <version_id>$version_id$</version_id> <!--user input, optional-->
```
rfrenchseti, line 17: Just a question: Do we want all these comments in the final label files?

pds4indextools/template_pds.xml
```diff
@@ -0,0 +1,178 @@
...
+            <doi>$Citation_Information['doi']$</doi> <!--user input, optional-->
+        $FOR(Citation_Information['keyword'])
+            <keyword>$VALUE$</keyword> <!--user input, optional-->
+        $END_FOR
+            <description>$Citation_Information['description']$</description> <!--user input, optional-->
+        $IF(Citation_Information.get('Funding_Acknowledgement']))
+            <Funding_Acknowledgement><!--user input, optional-->
+                <funding_source>$Funding_Acknowledgement['funding_source']$</funding_source> <!--user input, optional-->
+                <funding_year>$Funding_Acknowledgement['funding_year']$</funding_year> <!--user input, optional-->
+                <funding_award>$Funding_Acknowledgement['funding_year']$</funding_award> <!--user input, optional-->
```
rfrenchseti, line 35: Are these always the same year?

pds4indextools/template_pds.xml
```diff
@@ -0,0 +1,178 @@
...
+            <Modification_Detail>
+                <modification_date>$field['modification_date']$</modification_date> <!--user input, recommended-->
+                <version_id>$field['version_id']$</version_id>   <!--user input, recommended-->
+                <description>$field['description']$</description> <!--user input, recommended-->
+            </Modification_Detail>
+            $END_FOR
+        </Modification_History>
+     $END_IF
+        <License_Information>
+            <name>Creative Common Public License CC0 1.0 (2024)</name>
```
rfrenchseti, line 53: Are all PDS4 labels licensed this way? Would it ever make sense for a user to produce a proprietary label?

pds4indextools/template_pds.xml
```diff
@@ -0,0 +1,178 @@
...
+            <curating_facility></curating_facility>
+            <description></description>
+        </Source_Product_External>
+        $END_FOR
+        $END_IF
+    </Reference_List>
+    $IF(Product_Ancillary)
+    <File_Area_Ancillary>
+    $END_IF
+    $IF(Product_Metadata_Supplemental)
```
rfrenchseti, line 104: Another case where `$ELSE` would be appropriate, or a validation test?

test_files/expected/clean_header_field_names_success_2.csv
```diff
@@ -0,0 +1,3 @@
+pds_Product_Observational__pds_Identification_Area<1>__pds_logical_identifier<1>,pds_Product_Observational__pds_Identification_Area<1>__pds_version_id<1>,pds_Product_Observational__pds_Observation_Area<1>__pds_Time_Coordinates<1>__pds_start_date_time<1>,pds_Product_Observational__pds_Observation_Area<1>__pds_Time_Coordinates<1>__pds_stop_date_time<1>,pds_Product_Observational__pds_Observation_Area<1>__pds_Primary_Result_Summary<1>__pds_Science_Facets<1>__pds_wavelength_range<1>,pds_Product_Observational__pds_Observation_Area<1>__pds_Primary_Result_Summary<1>__pds_Science_Facets<2>__pds_wavelength_range<1>
```
rfrenchseti, line 1: If these are supposed to be clean field names, why do they still have `<1>`?

test_files/samples/element_extra_file_info_2.txt
```diff
@@ -0,0 +1,2 @@
+**/pds:logical_identifier*
```
rfrenchseti, line 1: Given the filename, I assumed this would include "extra file" elements like `filepath`.

tests/test_pds4_create_xml_index_2.py
```diff
@@ -0,0 +1,232 @@
+from lxml import etree
```
rfrenchseti, line 1: `test_pds4_create_xml_index.py`` appears to be "black box" tests - where we run the whole program without understanding how it works internally. `test_pds4_create_xml_index_2.py` appears to be "white box" tests, where we are calling individual functions and doing unit tests on them. I would rename the files to `test_pds4_create_xml_index_whitebox.py` etc.

rfrenchseti (changes requested):
A few comments on everything except the main program.

test_files/expected/clean_header_field_names_success_2.csv
esimpsons3ti: This file and `clean_header_field_names_success.csv` are actually deprecated, since they were replaced in the tests with their `.txt` counterparts. The `.txt` files are correct, and are actually being tested.

pds4indextools/template_pds.xml
esimpsons3ti: This was done on the advice of Mitch. I imagine if a proprietary label did need to be made, they could add in a `License_Information` section on their own. Currently the template does not create `License_Information` sections.

pds4indextools/template_pds.xml
esimpsons3ti: Ah, copy-paste typo.

pds4indextools/template_pds.xml
esimpsons3ti: I have removed them, since they really don't lend anything to the final product anyway.

pds4indextools/template_pds.xml
esimpsons3ti: I tried using $ELSE during development, and I got this version working faster than I got $ELSE working. Currently the command `--generate-label` requires you to choose either or, and sets that one to True.

test_files/expected/clean_header_field_names_success_2.csv
rfrenchseti: OK, then the old files that aren't being used should be deleted...

test_files/samples/element_extra_file_info_2.txt
esimpsons3ti: Because `filepath` has to be invoked at the command line, it is generated at the very end of the index file generation. 

I will change the filename to something else.

pds4indextools/template_pds.xml
esimpsons3ti: I am going to mark this as acknowledged. In the next round of changes to make I'll consider the above changes.

pds4indextools/template_pds.xml
matthewtiscareno: I don't understand the question, but it seems like the second instance of `funding_year` should be `funding_award`, so that it matches the attribute name. 

These are three attributes describing a single source of funding that is being acknowledged.  Three values for a single award might be NASA, 2024, and CDAP.

pds4indextools/template_pds.xml
matthewtiscareno: With rare exceptions, I expect CC0 to be the license specified in labels.  However, we should always allow for other values, because they may occur.  For example, some data from the CLPS program come from commercial providers and might not be CC0, though RMS is unlikely to deal with such data.

matthewtiscareno:
I ran the tool a couple days ago.  It initially didn't work, but @esimpsons3ti sat down with me and diagnosed the problem.  It was this line in the `process_schema_location` function:
`tree = etree.parse(file_path)`
For some reason, the way I configured the VSCode JSON was inputting file_path as a PosixPath rather than as a string, which broke the code.  We fixed it by simply changing that line so that it read
`tree = etree.parse(str(file_path))`
I'm not sure whether this fix, or a different fix for that situation, has been added to the branch. 

At this point I cannot run the tool at all because yesterday's changes seem to have introduced a requirement of `pdstemplate`.  I have the `rms-pdstemplate` repo on my machine, but I do not know how to get this tool to find it.  This might be an occasion for me to improve my usage of VSCode and maybe even virtual machines (which I do not currently use but probably should). 

More generally, is there documentation that we would expect a novice user to read that would enable them to run the program?  I see the docstring, but it doesn't address questions like what software must you have installed or what commands you must execute in order to be able to run the tool successfully.

esimpsons3ti:
@matthewtiscareno I understand that the docstrings within the program are not sufficient for a user to go off of. The current plan is to save creating the README file for last. However, I do intend on making a draft of it today.

tests/test_pds4_create_xml_index_blackbox.py
```diff
@@ -0,0 +1,397 @@
+from pathlib import Path
+import pytest
+import os
+import sys
+import tempfile
+sys.path.append(str(Path(__file__).resolve().parent.parent / "pds4indextools"))
```
rfrenchseti, line 6: Doesn't pytest put the current directory into PYTHONPATH? So then you could just import this as:
```
import pds4indextools.pds4_create_xml_index as tools
```
and avoid modifying the `sys.path`.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -1,69 +1,200 @@
...
+This script scrapes label files within specified directories, extracts information from
+user-defined XPaths/elements, and generates either an index file or a .txt file containing
+the XPaths available to the user. The script provides options for customizing the
+extraction process, such as:
+    - Specifying desired content by either limiting the results or requesting additional
+      file information.
+    - Sorting the resulting file by a user-specified value.
+    - Allowing for a user-made configuration file (.ini) for further custom content.
 
 Usage:
```
rfrenchseti, line 13: Honestly I would just delete all of this from the docstring and tell the user to run `python pds4_create_xml_index.py --help`. That way you don't have to keep those two things in sync, and the help should always have all the needed information anyway.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -75,61 +206,111 @@ def default_value_for_nil(config, data_type, nil_value):
...
+            Patterns starting with '!' indicate keys that should be excluded from the
+            result.
+        verboseprint (function): A function for printing verbose messages.
 
     Returns:
-        Converted XPath expression.
+        dict: A filtered dictionary containing only the keys that match the inclusion
+        patterns and do not match the exclusion patterns.
+
+    How it works:
```
rfrenchseti, line 249: This is not a valid docstring section. See https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html under "Docstring Sections" to see the valid section headers.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -1,69 +1,200 @@
...
+        str: Converted XPath expression.
+
+    Example:
+        >>> tree = etree.parse(str(path_to_label_file))
+        >>> root = tree.getroot()
+        >>> xml_header_path = '/*/*[1]/*[2]'
+        >>> namespaces = root.nsmap
+        >>> convert_header_to_xpath(root, xml_header_path, namespaces)
+        'pds:Product_Observational/pds:Identification_Area[1]/pds:version_id[2]'
+
```
rfrenchseti, line 117: Put the blank line after the triple-quote, not before. Or don't have a blank line at all, which seems to be your style elsewhere.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -336,133 +738,545 @@ def write_results_to_csv(results_list, args, output_csv_path):
...
-    config = load_config_file(args.config_file)
+    config = load_config_file(specified_config_file=args.config_file)
 
     directory_path = Path(args.directorypath)
-    patterns = args.pattern
+    verboseprint(f'Top level directory to scrape: {directory_path}')
+    patterns = args.patterns
+    verboseprint(f'Patterns to scrape: {patterns}')
+
+    """
```
rfrenchseti, line 932: Docstrings should not be randomly placed in the code. They always have to be attached to a particular variable, function, method, class, etc. I think here you want to use a block comment, like:
```
# Initializing the required arguments: directorypath and patterns. These
# will determine which files will be scraped for.

File: pds4indextools/pds4_create_xml_index.py [hunk 36]
```diff
@@ -336,133 +738,545 @@ def write_results_to_csv(results_list, args, output_csv_path):
...
-    verboseprint(f'{len(label_files)} matching files found')
+    verboseprint(f'{len(label_files)} matching file(s) found')
 
     if label_files == []:
-        print(f'No files matching {pattern} found in directory: {directory_path}')
+        print(f'No files matching any patterns found in directory: {directory_path}')
         sys.exit(1)
 
+    """
+    Loading in additional patterns from --elelemts-file, if applicable,
```
rfrenchseti, line 956: Spelling

esimpsons3ti:
@rfrenchseti Correct, there are currently no tests for label generation. That will be today's project, alongside general unit test coverage.

rfrenchseti:
I'm looking at the `--help` with the eyes of a brand new user.

- I'm finding these confusing:
```
  --output-csv-file OUTPUT_CSV_FILE
                        Specify the location and filename of the index file.
  --output-txt-file OUTPUT_TXT_FILE
                        Specify the location and filename of the XPath headers file.
```
Why is the option to specify the index file called --output-csv-file instead of --output-index-file? Why is the option to specify the headers file called --output-txt-file instead of --output-headers-file? Also the capitalized string should probably be something like FILENAME to make it clear what you specify as an argument. If you just duplicate the name of the option it's confusing.

- `--elements-file` doesn't really specify elements, right? An element is a pair of tags plus the thing between the tags, like `<a>value</a>`. What you're actually specifying is full XPaths to scrape. So maybe it should be called something like --limit-xpaths-file?

- Similar to above, `--sort-by SORT_BY` should say something useful instead of `SORT_BY`, like `HEADER_NAME`.

- In
```
  --add-extra-file-info {LID,filename,filepath,bundle_lid,bundle} [{LID,filename,filepath,bundle_lid,bundle} ...]
                        Add additional columns to the final index file. Choose from the following: "LID", "filename", "filepath", "bundle_lid", and "bundle". If specifying multiple column
                        names, supply them as separate arguments separated by spaces
```
you don't need the "Choose from" sentence, because the options are already listed above. Also there should be a period at the end.

- For
```
  --config-file CONFIG_FILE
                        Read a user-specified configuration file. File must be a .ini file.
```
It's not true the file has to be a .ini file. I think you want to say "The file must be in .ini file format."

- For
```
  --dump-available-xpaths
                        Give a .txt file of all XPaths within the given label files. This file can be used as a base file for --elements-file.
```
"Give" is ambiguous. Maybe say "Write". Does this go along with `--output-txt-file`? If so, it would be a lot easier to have a single command line option `--dump-available-xpaths FILENAME`. If it's present, do the dump to the given file. If it's not present, don't dump anything. Also, can you clarify whether these XPaths are before or after simplification? I think they need to be before, because otherwise you can't use them in the "elements" file if you have `--simplify-xpaths` turned on.

- For
```
  --generate-label {Product_Ancillary,Product_Metadata_Supplemental}
                        Generate a PDS4 label for the generated index file. Can generate either a Product_Ancillary or Product_Metadata_Supplemental label.
```
what is the filename of the resulting label?

- For
```
  --label-user-input LABEL_USER_INPUT
                        Provide an optional .yaml file containing additional information for the generated label.
```
again it doesn't have to be a file with the extension `.yaml`, it just has to be in YAML format.

- I would add an argparse `epilog` that refers people to the online documentation for more details.

- Why is there both a "config file" and a "label user input" file? Wouldn't it be easier to put the label info into the config file? Or do we expect the user to want to keep those separate?

- The order of the options as displayed is important to the user's understanding. In fact, I might group them together using argparse's "argument_group" concept. The groupings would be, in this order, with the options in this order (open to discussion):
  - Index file generation:
    - --output-csv-file
    - --add-extra-file-info
    - --sort-by
    - --fixed-width
    - --clean-header-field-names
    - --simplify-xpaths
  - Limiting XPaths:
    - --elements-file
    - --dump-available-xpaths
    - --output-txt-file
  - Label file generation:
    - --generate-label
    - --label-user-input
  - Miscellaneous:
    - --config-file
    - --verbose

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: I have replaced it with "Attention".

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: Huh, `flake8` didn't catch this one.

rfrenchseti:
Now that I look at the code, it seems like `--dump-available-xpaths` is not actually used for anything anymore and that you already did the change to make `--output-txt-file` do all the work. So you just have to remove the dead argument.

rfrenchseti:
I think in the `-add-extra-file-info` option, we shouldn't capitalize `LID` since we aren't capitalizing `bundle_lid`.

rfrenchseti:
If you specify both `--output-txt-file` and `--output-csv-file`, you only get the txt file. It silently fails to write the csv file, which is really confusing, especially if you already had a csv file there that you were trying to update (I spent the last 10 minutes trying to figure out why my csv file had the wrong contents). I see no reason that we can't allow writing both the csv and the header file in the same run of the program.

rfrenchseti:
`--simplify-xpaths` is changing the order of the header fields. It should keep them the same. See the attached two files as examples.
[rings.csv](https://github.com/user-attachments/files/16058961/rings.csv)
[rings_simp.csv](https://github.com/user-attachments/files/16058964/rings_simp.csv)

rfrenchseti:
Separating the arguments to `--add-extra-file-info` and `--sort-by` as separate arguments means that argparse will fail to parse the command line in some cases. For example:
```
❯ python pds4_create_xml_index.py --add-extra-file-info filepath filename LID /mnt/rms-holdings/pds4-holdings/bundles/uranus_occs_earthbased/uranus_occ_u0_kao_91cm "**/data/rings/*0m.xml" --verbose --generate-label Product_Ancillary
usage: pds4_create_xml_index.py [-h] [--output-csv-file OUTPUT_CSV_FILE] [--output-txt-file OUTPUT_TXT_FILE] [--elements-file ELEMENTS_FILE] [--simplify-xpaths] [--verbose]
                                [--sort-by SORT_BY [SORT_BY ...]] [--clean-header-field-names]
                                [--add-extra-file-info {LID,filename,filepath,bundle_lid,bundle} [{LID,filename,filepath,bundle_lid,bundle} ...]] [--config-file CONFIG_FILE]
                                [--dump-available-xpaths] [--fixed-width] [--generate-label {Product_Ancillary,Product_Metadata_Supplemental}] [--label-user-input LABEL_USER_INPUT]
                                directorypath patterns [patterns ...]
pds4_create_xml_index.py: error: argument --add-extra-file-info: invalid choice: '/mnt/rms-holdings/pds4-holdings/bundles/uranus_occs_earthbased/uranus_occ_u0_kao_91cm' (choose from 'LID', 'filename', 'filepath', 'bundle_lid', 'bundle')
```
This happens whenever one of the arguments that takes multiple values separated by spaces is followed by the required directory name and patterns. This will be confusing to the user, because we explicitly say the directory name and patterns should come last. In fact, I think they should be able to be provided in any order. The only fix for this is to change the behavior of `--add-extra-file-info` and `--sort-by`. I think they should each take a comma-separated list of items instead of a `+` set of arguments. Then just do a `split(',')` on the values to get the argument values. You'll have to write your own validation function now, since argparse won't be able to do it. But this is easy to do. Here's an example:

```
def _validate_planet(planet):
    """Routine for argparse to validate a planet."""
    if not planet in ("jupiter", "saturn", "uranus", "neptune"):
        raise argparse.ArgumentTypeError(
             f"{planet} is not a valid planet - must be one of "
              "jupiter, saturn, uranus, neptune")
    return planet

group.add_argument(
    "--planet", default="saturn",
    type=_validate_planet,
    help=f"""Which planet to process: jupiter, saturn, uranus, neptune
             (saturn is the default)""")
```

rfrenchseti:
If you don't specify any output path, you get the error:
```
Output path not chosen.
```

First of all, you can print an error for this long before you scrape all the labels. Right now it does all the label scraping (which can take a long time) and then complains. But the use of passive voice and "chosen" is confusing. I would suggest something like "You must specify at least one of: --output-csv-file or --output-txt-file" etc.

rfrenchseti:
Generating a label file is crashing for me:

```
❯ python pds4_create_xml_index.py /mnt/rms-holdings/pds4-holdings/bundles/uranus_occs_earthbased/uranus_occ_u0_kao_91cm "**/data/rings/*0m.xml" --generate-label Product_Ancillary --output-csv-file rings2.csv
  File "/seti/all_repos/rms-pds4indextools/pds4indextools/pds4_create_xml_index.py", line 1283, in <module>
    main()
  File "/seti/all_repos/rms-pds4indextools/pds4indextools/pds4_create_xml_index.py", line 1217, in main
    true_type = find_base_attribute(xsd_tree, name)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/seti/all_repos/rms-pds4indextools/pds4indextools/pds4_create_xml_index.py", line 783, in find_base_attribute
    target_attribute_value = target_attribute_values[0]
                             ~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

rfrenchseti:
When you dump the elements file with `--output-txt-file` while using `--simplify-xpaths`, you get the simplified XPaths twice, once in their simplified form and once in their original form. See the attached file. Example (excerpt from full file):

```
pds:logical_identifier
pds:title
pds:information_model_version
pds:Product_Observational/pds:Identification_Area<1>/pds:logical_identifier<1>
pds:Product_Observational/pds:Identification_Area<1>/pds:title<1>
pds:Product_Observational/pds:Identification_Area<1>/pds:information_model_version<1>
```

generated with:

```
python pds4_create_xml_index.py /mnt/rms-holdings/pds4-holdings/bundles/uranus_occs_earthbased/uranus_occ_u0_kao_91cm "**/data/rings/*0m.xml" --simplify-xpaths --output-txt-file rings2_simp.txt
```

[rings2_simp.txt](https://github.com/user-attachments/files/16059097/rings2_simp.txt)

matthewtiscareno:
> If you don't specify any output path, you get the error:
> 
> ```
> Output path not chosen.
> ```
> 
> First of all, you can print an error for this long before you scrape all the labels. Right now it does all the label scraping (which can take a long time) and then complains. But the use of passive voice and "chosen" is confusing. I would suggest something like "You must specify at least one of: --output-csv-file or --output-txt-file" etc.

If the user doesn't specify an output path, wouldn't it be better to write to a default output path, rather than for the tool to crash?

rfrenchseti:
> > If you don't specify any output path, you get the error:
> > ```
> > Output path not chosen.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > First of all, you can print an error for this long before you scrape all the labels. Right now it does all the label scraping (which can take a long time) and then complains. But the use of passive voice and "chosen" is confusing. I would suggest something like "You must specify at least one of: --output-csv-file or --output-txt-file" etc.
> 
> If the user doesn't specify an output path, wouldn't it be better to write to a default output path, rather than for the tool to crash?

Previously if you didn't specify an output file, it created a file with a name like `index.csv` in the root directory of the bundles/collections. I argued that this was a bad default, because it was writing into the (possibly read-only, and at least archival) directory structure and thus modifying something the user wasn't expecting to be modified. I think it would be perfectly fine to write into a default filename as long as 1) the filename was in the user's current directory, not the bundle root, 2) the program printed a message saying what filename it was writing and, and maybe 3) didn't overwrite a file that was already there.

matthewtiscareno:
> I think it would be perfectly fine to write into a default filename as long as 1) the filename was in the user's current directory, not the bundle root, 2) the program printed a message saying what filename it was writing and, and maybe 3) didn't overwrite a file that was already there.

I totally agree.  I might accomplish #3 by first checking whether `index.csv` already exists in the user's current directory.  If it does, then change the output name to `index1.csv`.  If that also already exists, then go on to `index2.csv`, and so forth.

esimpsons3ti:
> This will be confusing to the user, because we explicitly say the directory name and patterns should come last.

Where is that said? They're listed first in the examples we gave.

esimpsons3ti:
> Generating a label file is crashing for me:
> 
> ```
> ❯ python pds4_create_xml_index.py /mnt/rms-holdings/pds4-holdings/bundles/uranus_occs_earthbased/uranus_occ_u0_kao_91cm "**/data/rings/*0m.xml" --generate-label Product_Ancillary --output-csv-file rings2.csv
> Traceback (most recent call last):
>   File "/seti/all_repos/rms-pds4indextools/pds4indextools/pds4_create_xml_index.py", line 1283, in <module>
>     main()
>   File "/seti/all_repos/rms-pds4indextools/pds4indextools/pds4_create_xml_index.py", line 1217, in main
>     true_type = find_base_attribute(xsd_tree, name)
>                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/seti/all_repos/rms-pds4indextools/pds4indextools/pds4_create_xml_index.py", line 783, in find_base_attribute
>     target_attribute_value = target_attribute_values[0]
>                              ~~~~~~~~~~~~~~~~~~~~~~~^^^
> IndexError: list index out of range
> ```

This has been fixed with changes to `--simplify-xpath`'s behavior.

rfrenchseti:
> > This will be confusing to the user, because we explicitly say the directory name and patterns should come last.
> 
> Where is that said? They're listed first in the examples we gave.

It's in the `--help` output. I don't know why it's last there, since you put it first in your list of arguments, but I guess it just lists unnamed arguments at the end by default. Still, I think having multiple values be separated by spaces is confusing. That's not something command line programs normally do for multiple options - they have them separated by commas. If we just changed that convention, it would solve this problem entirely.

```
usage: pds4_create_xml_index.py [-h] [--output-csv-file OUTPUT_CSV_FILE] [--output-txt-file OUTPUT_TXT_FILE] [--elements-file ELEMENTS_FILE] [--simplify-xpaths] [--verbose]
                                [--sort-by SORT_BY [SORT_BY ...]] [--clean-header-field-names]
                                [--add-extra-file-info {LID,filename,filepath,bundle_lid,bundle} [{LID,filename,filepath,bundle_lid,bundle} ...]] [--config-file CONFIG_FILE]
                                [--dump-available-xpaths] [--fixed-width] [--generate-label {Product_Ancillary,Product_Metadata_Supplemental}] [--label-user-input LABEL_USER_INPUT]
                                directorypath patterns [patterns ...]
```

esimpsons3ti:
> Separating the arguments to `--add-extra-file-info` and `--sort-by` as separate arguments means that argparse will fail to parse the command line in some cases. For example:
> 
> ```
> ❯ python pds4_create_xml_index.py --add-extra-file-info filepath filename LID /mnt/rms-holdings/pds4-holdings/bundles/uranus_occs_earthbased/uranus_occ_u0_kao_91cm "**/data/rings/*0m.xml" --verbose --generate-label Product_Ancillary
> usage: pds4_create_xml_index.py [-h] [--output-csv-file OUTPUT_CSV_FILE] [--output-txt-file OUTPUT_TXT_FILE] [--elements-file ELEMENTS_FILE] [--simplify-xpaths] [--verbose]
>                                 [--sort-by SORT_BY [SORT_BY ...]] [--clean-header-field-names]
>                                 [--add-extra-file-info {LID,filename,filepath,bundle_lid,bundle} [{LID,filename,filepath,bundle_lid,bundle} ...]] [--config-file CONFIG_FILE]
>                                 [--dump-available-xpaths] [--fixed-width] [--generate-label {Product_Ancillary,Product_Metadata_Supplemental}] [--label-user-input LABEL_USER_INPUT]
>                                 directorypath patterns [patterns ...]
> pds4_create_xml_index.py: error: argument --add-extra-file-info: invalid choice: '/mnt/rms-holdings/pds4-holdings/bundles/uranus_occs_earthbased/uranus_occ_u0_kao_91cm' (choose from 'LID', 'filename', 'filepath', 'bundle_lid', 'bundle')
> ```
> 
> This happens whenever one of the arguments that takes multiple values separated by spaces is followed by the required directory name and patterns. This will be confusing to the user, because we explicitly say the directory name and patterns should come last. In fact, I think they should be able to be provided in any order. The only fix for this is to change the behavior of `--add-extra-file-info` and `--sort-by`. I think they should each take a comma-separated list of items instead of a `+` set of arguments. Then just do a `split(',')` on the values to get the argument values. You'll have to write your own validation function now, since argparse won't be able to do it. But this is easy to do. Here's an example:
> 
> ```
> def _validate_planet(planet):
>     """Routine for argparse to validate a planet."""
>     if not planet in ("jupiter", "saturn", "uranus", "neptune"):
>         raise argparse.ArgumentTypeError(
>              f"{planet} is not a valid planet - must be one of "
>               "jupiter, saturn, uranus, neptune")
>     return planet
> 
> group.add_argument(
>     "--planet", default="saturn",
>     type=_validate_planet,
>     help=f"""Which planet to process: jupiter, saturn, uranus, neptune
>              (saturn is the default)""")
> ```

Currently, `--sort-by` can sort by anything, not just the values in `--add-extra-file-info`. Do we want to limit `--sort-by` to only sort by the allowed values for `--add-extra-file-info`, or continue to allow users to sort by anything?

rfrenchseti:
> Currently, `--sort-by` can sort by anything, not just the values in `--add-extra-file-info`. Do we want to limit `--sort-by` to only sort by the allowed values for `--add-extra-file-info`, or continue to allow users to sort by anything?

I'm not sure where this question is coming from. Why would we change the behavior of `--sort-by`? The only thing I'm saying is that instead of: `--sort-by field1 field2 field3` we require `--sort-by field1,field2,field3`. Just take a single string argument instead of "+", and split on commas to get the list of fields. For the additional file info fields, you'll have to validate the entries in that new list.

esimpsons3ti:
Note: Unit test for label generation have not been added yet. I'm just making a new commit to contain the changes from the above comments.

pds4indextools/pds4_create_xml_index.py
rfrenchseti: This comment is marked as checked off but all the usage data is still there.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -25,45 +31,118 @@
...
+    --dump-available-xpaths
+                         Create a .txt file containing all available XPath headers for
+                         given label file(s). Can be modified and used as a file for
+                         --elements-file
+    --fixed-width        Create a fixed-width index file. 
 
 Example:
-python3 pds4_create_xml_index.py <toplevel_directory> "glob_path1" "glob_path2" 
---output_file <outputfile> --elements-file sample_elements.txt --verbose
+    python3 pds4_create_xml_index.py <toplevel_directory> "glob_path1" "glob_path2"
```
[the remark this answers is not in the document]
rfrenchseti: You can change it later, but eventually the proper location for documentation (in the postscript of the help message) will be `https://rms-pds4indextools.readthedocs.io/en/latest`

pds4indextools/pds4_create_xml_index.py
```diff
@@ -198,271 +701,781 @@ def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
...
-
-    # Filter XSD URLs
-    xsd_urls = [url for url in schema_location_values if url.endswith('.xsd')]
+    def pad_column_values_and_headers(df):
+        """
+        Pad the values and headers of a DataFrame to align column widths.
 
-    return xsd_urls
+        Args:
+        df (pandas.DataFrame): The DataFrame whose column values and headers need to be
```
rfrenchseti, line 719: Should this and the Returns below be indented an extra level?

pds4indextools/pds4_create_xml_index.py
```diff
@@ -198,271 +701,781 @@ def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
...
+
+    return None
+
+
+def scrape_namespaces(xsd_url):
+    """
+    Fetch and parse an XSD file from a given URL to extract namespace declarations.
+
+    Parameters:
+    xsd_url (str): The URL of the XSD file to be fetched and parsed.
```
rfrenchseti, line 905: It seems like a lot of these new docstrings don't have indents after the Parameters and Returns and Raises. They should.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -198,271 +701,781 @@ def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
...
+            if 'ASCII' in base_type or 'UTF8' in base_type:
+                return base_type
+
+            next_query = (
+                f".//xs:simpleType[@name='{base_type.split(':')[-1]}']"
+                f"//xs:restriction/@base"
+            )
+            try:
+                next_result = xsd_tree.xpath(next_query, namespaces=namespaces)
+            except Exception:
```
rfrenchseti, line 812: "Naked" exceptions are generally a bad idea. Do you know which exceptions you want to catch here?

pds4indextools/pds4_create_xml_index.py
```diff
@@ -198,271 +701,781 @@ def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
...
+    # Define the initial XPath query to find the complexType element by name
+    initial_query = f".//xs:complexType[@name='{target_name}']//xs:extension/@base"
+    initial_result = get_base_type(initial_query)
+
+    if initial_result:
+        base_type = initial_result[0]
+        return follow_base_type(base_type)
+
+    # If the initial query did not find a result, try other patterns
+    other_queries = [
```
rfrenchseti, line 844: Why are these separated out? Can't you just put the initial query first in this list?

pds4indextools/pds4_create_xml_index.py
```diff
@@ -198,271 +701,781 @@ def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
...
+
+    # Return the creation date in ISO 8601 format
+    return dt_object.isoformat()
+
+
+def load_yaml_file(yaml_file):
+    """
+    Load and parse a YAML file.
+
+    Parameters:
```
rfrenchseti, line 961: Also it seems like you switched from "Args:" to "Parameters:". Mark has made the argument that "Parameters:" works better because it agrees with the way Sphinx formats HTML documentation, so feel free to go ahead and change your old "Args:" to "Parameters:".

pds4indextools/pds4_create_xml_index.py
```diff
@@ -198,271 +701,781 @@ def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
...
+
+    Returns:
+    int: The length of the longest row in the CSV file, measured by the sum of the lengths
+         of the fields plus the number of delimiters in the row.
+
+    Raises:
+    FileNotFoundError: If the specified file does not exist.
+    """
+    with open(filename, 'r') as csvfile_fp:
+        reader = csv.reader(csvfile_fp, delimiter=',')  # Adjust the delimiter as needed
```
rfrenchseti, line 986: I don't understand the comment about the delimiter. Who are you saying should adjust it?

pds4indextools/pds4_create_xml_index.py
```diff
@@ -198,271 +701,781 @@ def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
...
+        longest_row_length = 0
+
+        for row in reader:
+            current_row_length = sum(len(field.strip()) for field in row) + len(row) - 1
```
rfrenchseti, line 990: Just curious - why do you do this instead of just reading the file directly and looking for the longest line? Is it necessary to strip the fields?

pds4indextools/pds4_create_xml_index.py
```diff
@@ -198,271 +701,781 @@ def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
...
+                                            'like *, ?, and **. If supplying multiple '
+                                            'patterns, separate with spaces. Surround '
+                                            'each pattern with quotes.')
+
+    index_file_generation.add_argument('--output-index-file', type=str,
+                                       help='Specify the location and filename of the '
+                                            'index file.')
+
+    index_file_generation.add_argument(
+        '--add-extra-file-info',
```
rfrenchseti, line 1071: One more nice thing you can do with argparse is add the `metavar` option. This is what shows up in `--help` instead of the option name. For example, currently we get:
```
  --add-extra-file-info ADD_EXTRA_FILE_INFO
                        Add additional columns to the final index file. If specifying multiple column names, supply them as separate arguments separated by spaces.
```
but "ADD_EXTRA_FILE_INFO" is not really helpful, and nowhere does it say what the options are. So you can add to the `add_argument` definition something like:
```
metavar='COMMA_SEPARATED_COLUMN_NAMES'
```
and then put in the help string something about what the valid options are.

I would do this for everything that takes an argument. For example:
```
--sort-by SORT_BY
```
would make more sense as:
```
--sort-by COMMA_SEPARATED_HEADER_NAME(s)
```
and
```
--generate-label GENERATE_LABEL
```
would make more sense as:
```
--generate-label {Product_Ancillary,Product_Metadata_Supplemental}
```

pds4indextools/pds4_create_xml_index.py
```diff
@@ -198,271 +701,781 @@ def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
...
+                                       'the label files are included.')
+
+    limiting_results.add_argument('--output-headers-file', type=str,
+                                  help='Specify the location and filename of the XPath '
+                                       'headers file.')
+
+    label_generation = parser.add_argument_group('Label Generation')
+    label_generation.add_argument('--generate-label', type=str, nargs=1,
+                                  help='Generate a PDS4 label for the generated index '
+                                       'file called "<output_index_file>_label.xml". '
```
rfrenchseti, line 1112: Why is there a `_label` on the filename? Normally the label for a file is the same name as the data file, just with an `.xml` extension.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -198,271 +701,781 @@ def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
...
+        correct_duplicates(label_results)
+
+        # The label_results dictionary will now be filtered according to the contents
+        # of the --limit-xpaths-file input file. If this command is not used, the original
+        # dictionary will be returned. Glob patterns are processed sequentially, with the
+        # first pattern having the highest priority.
+        label_results = filter_dict_by_glob_patterns(
+            label_results, elements_to_scrape, verboseprint)
+
+        if label_results == []:
```
rfrenchseti, line 1268: I'm always more of a fan of `len(label_results) == 0`. The reason is that if you say `[]` explicitly, if in the future you changed `filter_dict_by_glob_patterns` to return a tuple or something else, which wouldn't otherwise break the code, it *would* break this test since it isn't a list anymore. Of course the pure Python way to do it is `if not label_results:`

pds4indextools/index_label_template_pds.xml
```diff
@@ -0,0 +1,178 @@
...
+            <description></description>                                           
+            <Record_Character>
+                <fields>$fields$</fields> 
+                <groups>0</groups>
+                <record_length unit="byte"></record_length>                        
+                $FOR(field, k=Field_Content)
+                    <Field_Character>
+                        <name>$field['name']$</name>
+                        <field_number>$k + 1$</field_number>
+                        <field_location unit="byte">$field['field_location']$</field_location>
```
rfrenchseti, line 135: I think `field_location` and `field_length` need to be omitted if the table is not fixed-width.

rfrenchseti (changes requested):
- When creating a file using the default name like `index.csv` you should print a message like: `Writing index file to index.csv` so the user knows where to find it. You could just print this message all the time, not just when it's the default name, to make it consistent. Then you could add similar messages for the headers and the label and the XPaths file writes.

-

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: Leftover debugging comment.

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: This was when I was trying to figure out how to manage line lengths for fixed-width fields. With a fresh pair of eyes, I now know that this is far too complicated.

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: Easier to distinguish on first glance. I'll get rid of it.

pds4indextools/index_label_template_pds.xml
esimpsons3ti: There is a check during label generation whether args.fixed_width was used. If it was, it uses Field_Character. If it was not used, it will use Field_Delimited, which excludes these values.

rfrenchseti: Well, it's not working. I'm not using fixed-width and I'm getting `Field_Character` in my label.

esimpsons3ti: @rfrenchseti I found the issue. The template wasn't the problem. It was the default_values.yaml file. Table_Character was set to True for some reason.

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: For some reason, a single unit test is failing with this change. And I really can't tell why. `label_results` will have content in it, but it triggers the 'No results found: glob pattern(s) excluded all matches' message.

rfrenchseti: That's a case where you should dig deeper to see what the problem is. Unexplained behavior can point to a hidden bug. Don't just go back to the code that "works" if you don't know why the other doesn't work. Try printing out the content of label_results just before this test and then seeing what the output is when running the tests.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -198,271 +657,816 @@ def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
...
+                print(f'Index file {index_file} is not a CSV or tab-separated file.')
+                sys.exit()
+            reader = csv.reader(index_fp, delimiter=delimiter)
+            headers = next(reader)
+
+            offset = 0
+            field_number = 0
+            jump = len(delimiter)
+            field_location = 1
+            maximum_field_lengths = max_field_lengths(index_file)
```
rfrenchseti, line 1385: The way to avoid this awkward use of max vs. maximum is to make the function have a verb in it: `compute_max_field_lengths`.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -198,271 +657,816 @@ def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
...
+
+            # Each header is processed for it's information, which is then put into
```
rfrenchseti, line 1387: it's => its

pds4indextools/pds4_create_xml_index.py
```diff
@@ -198,271 +657,816 @@ def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
...
+            # header_info to be referenced later. Not all information will be put into
+            # the generated label, since some information depends on whether the index
+            # file is fixed-width or delimited.
+            for header in headers:
+                if header.strip() in valid_add_extra_file_info and 'lid' in header:
```
rfrenchseti, line 1392: Probably easier just to say `header = header.strip()` right after the for and then you don't need to do it over and over again inside the loop.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -198,271 +657,816 @@ def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
...
+        with open(index_file, 'r', encoding='utf-8') as index_fp:
+            full_header = index_fp.readline()
+            full_header_length = len(full_header)
+            try:
+                sample_data = index_fp.read(5000)
+                delimiter = sniffer.sniff(sample_data).delimiter
+                index_fp.seek(0)  # Reset file pointer to the beginning
+            except csv.Error:
+                print(f'Index file {index_file} is not a CSV or tab-separated file.')
+                sys.exit()
```
rfrenchseti, line 1377: Should have a return code. It defaults to zero, which means no error.

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: I decided to make a new variable, `stripped_header`, since the presence of whitespace has significance in some calculations for generated labels from fixed-width index files.

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: I also added a docstring.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -198,271 +657,816 @@ def update_nillable_elements_from_xsd_file(xsd_file, nillable_elements_info):
...
+
+            # Initialize the dictionary with headers and set their max length to 0
+            for header in reader.fieldnames:
+                max_lengths[header] = 0
+
+            # Iterate through each row to calculate maximum field lengths
+            for row in reader:
+                for header in reader.fieldnames:
+                    field_length = len(row[header])
+                    if field_length > max_lengths[header]:
```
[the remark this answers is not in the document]
esimpsons3ti: I reworked this logic.

pds4indextools/pds4_create_xml_index.py
esimpsons3ti: This has been changed again, to further simplify it. There is now `whole_header`, which preserves headers with additional whitespace

rfrenchseti (approved):

rfrenchseti:
I am approving this pull, with the understanding that there are additional issues that need to be addressed, including:
- #18 
- #17 
- #16

PR #26, opened 2024-07-09
Fixing minor bugs, changing `scrape_namespaces` for timeout issue 

### LINKED ISSUE #22: Unknown sort key causes crash
In `pds4_create_xml_index`, if the `--sort-by` option is specified with an unknown column name, an error is generated such as:

```
  File "/seti/all_repos/rms-pds4indextools/pds4indextools/pds4_create_xml_index.py", line 1529, in <module>
    main()
  File "/seti/all_repos/rms-pds4indextools/pds4indextools/pds4_create_xml_index.py", line 1368, in main
    write_results_to_csv(all_results, args, output_csv_path)
  File "/seti/all_repos/rms-pds4indextools/pds4indextools/pds4_create_xml_index.py", line 718, in write_results_to_csv
    df.sort_values(by=sort_values, inplace=True)
  File "/seti/all_repos/rms-pds4indextools/venv/lib/python3.12/site-packages/pandas/core/frame.py", line 7189, in sort_values
    k = self._get_label_or_level_values(by[0], axis=axis)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/seti/all_repos/rms-pds4indextools/venv/lib/python3.12/site-packages/pandas/core/generic.py", line 1911, in _get_label_or_level_values
    raise KeyError(key)
KeyError: 'is_instrument_host'
```

This is unfriendly to the user. Instead an error message such as:

```
Unknown sort key "is_instrument_host". For a list of available sort keys, use the --output-headers-file option.
```

would be preferable.

esimpsons3ti:
Addressed a few fixes:

1. If a user attempts to sort by a key that does not exist within the available headers, a new error message will now print:
```
Unknown sort key 'pds:lohical_identifier'. For a list of available sort keys, use the --output-headers-file option.
```

2. scrape_namespaces has been changed to use the Element trees of the schema files, instead of loading in the schema files every time it is used. This previously called a timeout error.

3. Fixed an issue where empty classes in label files were mistaken for attributes. Since classes don't have data types, any classes that are caught this way will be given the `inapplicable` value for their data type in label generation.

Fixes #22


pds4indextools/pds4_create_xml_index.py
```diff
@@ -715,7 +717,11 @@ def pad_column_values_and_headers(df):
 
     if args.sort_by:
         sort_values = str(args.sort_by).split(',')
-        df.sort_values(by=sort_values, inplace=True)
+        try:
+            df.sort_values(by=sort_values, inplace=True)
+        except KeyError as bad_sort:
+            print(f'Unknown sort key {bad_sort}. For a list of available sort keys, use '
+                  f'the --output-headers-file option.')
```
rfrenchseti, line 724: Need to exit the program after this error

esimpsons3ti: Does it? In my mind it might be a bit punishing to exit the program and not create an index file because they specified a bad sort key. In this case they can still get their index file, it's just unsorted.

rfrenchseti: Expecting a sorted file and not getting one should be a fatal error. People might miss the error message, or they might be running this in an automated script and thus never know the sort failed, producing the wrong index file results. A bad index file could make it all the way to public release if this error wasn't caught. If the user didn't care about the sort order that much, then they wouldn't have asked for it to be sorted...

rfrenchseti (approved):

PR #27, opened 2024-07-09
Preparing documentation for 1.0.0 release

esimpsons3ti:
I have removed the FIXIT statements within pds4_create_xml_index.rst file. These issues will be handled after the 1.0.0 release on Thursday.

The following content was removed:
```
(FIXIT) ADDITIONAL THINGS TO DISCUSS
Explanation of how we format XPaths (e.g. what do the <1>-style suffixes mean?)
Format and contents of the --config-file config file including discussion of nillable elements.
Format and contents of the --label-user-file config file (and should these be the same thing?)
```


esimpsons3ti:
This isn't right. That's way too many commits.

PR #30, opened 2024-07-15
Fixing documentation, adding explanatory comments to unit tests

### LINKED ISSUE #19: Add comments to unit tests clarifying the related executable command
In order to make the unit tests more human-readable, please add some additional explanatory comments of each entry. 

For example, the following currently occurs within :
```
            # Testing --limit-xpaths-file
            (
                str(expected_dir / 'limit_xpaths_file_success_1.txt'),
                'limit_xpaths_file.txt',
                [
                    str(test_files_dir),
                    str(labels_dir.name / Path('tester_label_1.xml')),
                    '--limit-xpaths-file',
                    str(samples_dir / 'element_1.txt'),
                    '--output-headers-file'
                ]
            ),
```

The suggestion here is to add the following as a second comment line, after `# Testing --limit-xpaths-file`:
```
            # Executable command: pds4_create_xml_index.py test_files test_files/labels/tester_label_1.xml --limit-xpaths-file test_files/samples/element_1.txt --output-headers-file
            # Compare result to golden copy: test_files/expected/limit_xpaths_file_success_1.txt
```

However, that's not quite right, because I just tried executing it and it didn't work.  Please correct it and then use that.

esimpsons3ti:
The documentation previously stated that an understanding of XPath was necessary to use the tool effectively. This has been replaced with a passage that instead demonstrates how to read and understand the XPath-based headers the tool creates. 

The blackbox unit tests have been given explanatory block comments for each entry. Each block comment contains the equivalent command-line query and the golden copy file to compare the results against.

Fixes #19

### DIFF
--- a/docs/pds4_create_xml_index.rst
+++ b/docs/pds4_create_xml_index.rst
@@ -18,34 +18,45 @@ and accessing structured data within PDS4-compliant datasets.
 XPath Syntax and Structure
 --------------------------
 
-Before using the tool, it is imperative that the user becomes comfortable with
-the XPath language and how it is parsed with ``lxml``.
+This tool generates headers that represent the position of each element in the XML
+hierarchy in a manner similar to `the standard XPath format <https://developer.mozilla.org/en-US/docs/Web/XPath>`_. 
+While familiarity with XPath syntax is beneficial, it is not necessary to use this tool
+effectively. Note that for simplicity we call our headers XPaths throughout this document
+and within the command line arguments, despite some minor syntax differences. Below we
+describe the syntax we use for our headers. An XPath header is read from left to right,
+starting with the root element and moving through each subsequent child element. Each
+element in the path is separated by a forward slash (/). If an element has multiple
+instances within a single parent, predicates (numbers within angle brackets) are used to
+specify the exact instance, such as ../pds:Observation_Area<1>/pds:version_id<1>, which
+selects the first version_id element in Observation_Area. The predicate numbers count the
+instances of the child element, rather than its structural position in the file. For
+example, given the XML fragment:
 
-When elements are scraped from a label, they are returned in a format akin to a
-filepath. The absolute XPath contains all parent elements up to the root element
-of the label. Each element after a certain depth can also contain predicates:
-numbers surrounded by square brackets. These predicates give information on the
-location of the element in the label file, in relation to surrounding elements.
+```
+<Product_Observational>
+    <Identification_Area>
+        <logical_identifier>...</logical_identifier>
+        <Citation_Information>
+            <author_list>...</author_list>
+            <publication_year>...</publication_year>
+            <keyword>...</keyword>
+            <keyword>...</keyword>
+            <description>...</description>
+        </Citation_Information>
+    </Identification_Area>
+</Product_Observational>
+```
 
-However, this module returns XPath headers that have been reformatted from XPath
-syntax.
+the available XPaths are:
 
-- Namespaces are replaced by their prefixes. Namespaces are URIs that identify
-  which schema an element belongs to. For readability, the full URI's are
-  replaced by their prefixes.
-
-- Square brackets are replaced by angled brackets. This module utilizes ``glob``
-  syntax for finding label files and filtering out elements/attributes. Because
-  square brackets have meaning within ``glob`` statements, they needed to be
-  replaced with angled brackets.
-
-- The values within the predicates have been renumbered. In XPath syntax,
-  predicates are used to determine the location of an element in relation to its
-  parent. While this is useful in other applications, this logic fails if
-  multiples of the element and its parent appear within the document. Even if
-  the elements all have different values, all of their XPaths would be the same.
-  Instead, the predicates are renumbered to reflect which instance of the
-  element the value is represented by.
+```
+Product_Observational<1>/Identification_Area<1>/logical_identifier<1>
+Product_Observational<1>/Identification_Area<1>/Citation_Information<1>/author_list<1>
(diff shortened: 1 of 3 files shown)


docs/pds4_create_xml_index.rst
```diff
@@ -18,34 +18,45 @@ and accessing structured data within PDS4-compliant datasets.
...
-  multiples of the element and its parent appear within the document. Even if
-  the elements all have different values, all of their XPaths would be the same.
-  Instead, the predicates are renumbered to reflect which instance of the
-  element the value is represented by.
+This tool generates headers that represent the position of each element in the XML
+hierarchy in a manner similar to `the standard XPath format <https://developer.mozilla.org/en-US/docs/Web/XPath>`_. 
+While familiarity with XPath syntax is beneficial, it is not necessary to use this tool
+effectively. Note that for simplicity we call our headers XPaths throughout this document
+and within the command line arguments, despite some minor syntax differences. Below we
+describe the syntax we use for our headers. An XPath header is read from left to right,
```
rfrenchseti, line 26: I think you need to start a new paragraph before "An XPath"

docs/pds4_create_xml_index.rst
```diff
@@ -18,34 +18,45 @@ and accessing structured data within PDS4-compliant datasets.
...
+starting with the root element and moving through each subsequent child element. Each
+element in the path is separated by a forward slash (/). If an element has multiple
```
rfrenchseti, line 28: Just say "Elements are separated". Maybe also put the / in double-back-ticks.

docs/pds4_create_xml_index.rst
```diff
@@ -18,34 +18,45 @@ and accessing structured data within PDS4-compliant datasets.
...
+instances within a single parent, predicates (numbers within angle brackets) are used to
+specify the exact instance, such as ../pds:Observation_Area<1>/pds:version_id<1>, which
```
rfrenchseti, line 30: These paths should be in double-back-ticks. Also version_id and Observation_Area.

docs/pds4_create_xml_index.rst
```diff
@@ -185,4 +197,9 @@ Miscellaneous
   be useful for debugging.
 
 - ``--config-file``: Specify a ``.ini``-style configuration file for further customization
-  of the extraction process.
+  of the extraction process. This file allows you to replace the field entry for any
+  element's data type. Its primary purpose is to handle nilled elements. Nilled elements
+  are elements that are intentionally omitted due to inapplicable, missing, unknown, or
+  anticipated values. The default configuration file (``pds4indextools.ini``) covers a
```
rfrenchseti, line 203: People who installed this tool don't have access to the default file so this isn't helpful. Also saying it covers "a specific set" isn't helpful since they don't know what that set is.

tests/test_pds4_create_xml_index_blackbox.py
```diff
@@ -18,6 +18,12 @@
         'golden_file,new_file,cmd_line',
         [
             # Testing --limit-xpaths-file
+            # Executable command: pds4_create_xml_index ../test_files/labels
```
rfrenchseti, line 21: I think the idea was that all of these examples would have the commands on a single line. That way you could cut-and-paste the command and run it yourself. Having them broken across multiple lines means you can't easily do this. It's OK if it makes the line too long. We can add an exception in `.flake8`.

docs/pds4_create_xml_index.rst
```diff
@@ -175,8 +186,9 @@ Label generation
   ``ancillary`` or ``Product_Metadata_Supplemental`` for ``supplemental``. Additional
   customization of the label can be provided with ``--label-user-input``.
 
-- ``--label-user-info``: Provide a file containing customization of the generated label.
-  The file must be in YAML format.
+- ``--label-user-input``: Provide a file containing customization of the generated label.
```
rfrenchseti, line 189: The next big documentation sections to write are going to be on the format of this file and the config file, once we decide on whether they are two files or one. We should make this decision soon before people starting writing their own files and then need to convert.

rfrenchseti (changes requested):

tests/test_pds4_create_xml_index_blackbox.py
esimpsons3ti: Something like this?
```
per-file-ignores =
    # line too long
    tests/pds4_create_xml_index_blackbox.py: E501
```

rfrenchseti: Yes

docs/pds4_create_xml_index.rst
```diff
@@ -18,34 +18,47 @@ and accessing structured data within PDS4-compliant datasets.
...
+While familiarity with XPath syntax is beneficial, it is not necessary to use this tool
+effectively. Note that for simplicity we call our headers XPaths throughout this document
+and within the command line arguments, despite some minor syntax differences. Below we
+describe the syntax we use for our headers. 
+
+An XPath header is read from left to right, starting with the root element and moving 
+through each subsequent child element. Elements are separated by a forward slash (``/``).
+If an element has multiple instances within a single parent, predicates (numbers within
+angle brackets) are used to specify the exact instance, such as
+``../pds:Observation_Area<1>/pds:version_id<1>``, which selects the first ``version_id``
```
rfrenchseti, line 32: Small nit but I would put three dots here. Two dots usually means "up one directory" while three dots is the English "ellipsis" which means "something is omitted here". You can save that for the next revision.

rfrenchseti (approved):

PR #32, opened 2024-07-18
Fixing pattern matching for `--limit-xpaths-file`

### LINKED ISSUE #23: --limit-xpaths-file isn't really in glob format
In `pds4_create_xml_index` the current code uses Python's `fnmatch` to match wildcard characters with lines in the `--limit-xpaths-file` file. However, `fnmatch` doesn't actually implement the `**` concept provided by `glob`, even though we claim it does in the documentation. Instead it treats each string as just a filename, and thus `*` reaches across "directories". As a result, the XPath `/a/b/c` would be matched by `*`, but it should only be matched by `*/*/*` or `**` or `**/*`. To fix this we need to implement our own version of `glob` that breaks the XPaths into sections at `/` boundaries and calls `fnmatch` on each section recursively. The code to do this can be stolen from the `glob` source code and modified appropriately.

https://github.com/python/cpython/blob/3.12/Lib/glob.py

esimpsons3ti:
Due to the use of `fnmatch` in the `filter_dict_by_glob_patterns` function, any use of `**` was considered two separate `*`, and not as the recursive wildcard `**`. A new helper function has been added to simulate the "match zero or more directories" logic to the pattern matching function. I also fixed a unit test file to more accurately find the intended pattern, as well as fixed a typo in a docstring.

Fixes #23

### DIFF
--- a/pds4indextools/pds4_create_xml_index.py
+++ b/pds4indextools/pds4_create_xml_index.py
@@ -98,7 +98,7 @@ def correct_duplicates(label_results):
 
     Example:
             # XPaths in label_results shortened for readability
-        >>> keys = list(label_result.keys())
+        >>> keys = list(label_result)
         >>> keys = [
                 ../geom:SPICE_Kernel_Identification<1>/geom:kernel_type<1>,
                 ../geom:SPICE_Kernel_Identification<1>/geom:kernel_type_1<1>,
@@ -107,7 +107,7 @@ def correct_duplicates(label_results):
                 ../geom:SPICE_Kernel_Identification<1>/geom:kernel_type_4<1>
                 ]
         >>> correct_duplicate(label_results)
-        >>> keys = list(label_result.keys())
+        >>> keys = list(label_result)
         >>> keys = [
                 ../geom:SPICE_Kernel_Identification<1>/geom:kernel_type<1>,
                 ../geom:SPICE_Kernel_Identification<2>/geom:kernel_type<1>,
@@ -117,7 +117,7 @@ def correct_duplicates(label_results):
                 ]
     """
     element_names = set()
-    for key in list(label_results.keys()):
+    for key in list(label_results):
         tag = key.split('/')[-1].split('<')[0]
         number = tag.split('_')[-1]
         if number.isdigit():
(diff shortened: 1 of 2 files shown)


pds4indextools/pds4_create_xml_index.py
```diff
@@ -183,6 +183,48 @@ def extract_logical_identifier(tree):
...
+        elif segments:
+            return match_segment(segments[0], pattern) and match_recursive_helper(
+                segments[1:], patterns[1:])
+        else:
+            return False
+
+    pattern_segments = pattern.split('/')
+    matched_keys = []
+
+    for key in data.keys():
```
rfrenchseti, line 220: I missed this on previous code reviews, but every place in the code you can just delete `.keys()`. A basic dict is an iterator that returns the key names. It's more efficient to just use the iterator instead of calling `.keys()` to create a new view of the keys.

pds4indextools/pds4_create_xml_index.py
```diff
@@ -183,6 +183,48 @@ def extract_logical_identifier(tree):
...
+        if not patterns:
+            return not segments
+
+        pattern = patterns[0]
+        if pattern == '**':
+            if match_recursive_helper(segments, patterns[1:]):
+                return True
+            return bool(segments) and match_recursive_helper(segments[1:], patterns)
+        elif segments:
+            return match_segment(segments[0], pattern) and match_recursive_helper(
```
rfrenchseti, line 212: This would format better as something like:
```
            return (match_segment(segments[0], pattern) and
                    match_recursive_helper(segments[1:], patterns[1:])
```

rfrenchseti (changes requested):
I don't see any immediate problem, but I'm not convinced the testing is thorough given the complexity of the matching function's capabilities.

This would be a good function to add to your whitebox testing. It would be easier to test there than just adding a lot of new command line executions. You should test all the corner cases like:
- data is empty
- pattern is the empty string
- `**` is used to match zero directories, like `/a/**/c` matches `/a/c`.
- `**` is used to match some, but not all directories, like `/a/**/d/e` matches `/a/b/c/d/e`.
- `**` is used to match the final filename, like `/a/b/**` matches `/a/b/c`
- `/a/**/d/e` doesn't match `/a/b/c/e`
- You can also include cases with just `*` and `[seq]` and `[!seq]`
- Etc

In other words, write a bunch of tests with the intent of breaking the code with odd cases.

esimpsons3ti:
I was planning on working on unit tests after this is resolved, anyway. I would like that to be a part of a different pull request.

rfrenchseti (approved):

PR #33, opened 2024-07-23
New Indexing Tool: `pds4_create_collection_product`

esimpsons3ti:
This pull contains:
- A new indexing tool: `pds4_create_collection_product`. This tool creates collection products for a given collection. 
- A new rST file of documentation for this new tool
- A modified `pyproject.toml` file to include this new tool.


rfrenchseti:
For future reference, you can deploy new code to the test version of PyPI so that you can do a trial install using pip and make sure things work as expected. To do this:
- Go to the "Actions" tab on https://github.com/SETI/rms-pds4indextools
- Select "Publish to Test PyPI" on the left. Make sure it's the test version!
- On the right is a dropdown that says "Run workflow". Click it, then select the branch you want to test, and click the green "Run workflow"
- After the action completes, check out https://test.pypi.org/project/rms-pds4indextools/
- You should see "Released:" and some recent time at the top
- Also at the top is the pip command to execute to install this version, which will look something like `pip install -i https://test.pypi.org/simple/ rms-pds4indextools`
You can install it into a fresh venv just like you would normally and test it to see if it works.

docs/pds4_create_collection_product.rst
```diff
@@ -0,0 +1,52 @@
...
+
+
+Optional arguments
+^^^^^^^^^^^^^^^^^^
+
+``--collection-product-file COLLECTION_PRODUCT_FILEPATH``: Specify the location and name
+of the collection product. This allows for the collection product to be generated outside
+of the collection it represents. It is recommended that the file have the suffix ``.csv``.
+If no directory is specified, the collection product will be written to the current
+directory. If this command is not used, the collection product will be written to the path
```
rfrenchseti, line 51: "If this argument is not specified, the collection product will be written to the top-level directory of the collection with the name collection_<dirname>.csv" or something like that

docs/pds4_create_collection_product.rst
```diff
@@ -0,0 +1,52 @@
+``pds4_create_collection_product`` Program
+=================================
+
+Introduction
+------------
+The RMS Node's PDS4 Collection Product Creation Tool (``pds4_create_collection_product``)
+is designed to create bespoke collection product files from
+`Planetary Data System (PDS) <https://pds.nasa.gov>`_ `PDS4-format
+<https://pds.nasa.gov/datastandards/documents/>`_label files. This tool automates the
```
rfrenchseti, line 9: Add space after _

docs/pds4_create_collection_product.rst
```diff
@@ -0,0 +1,52 @@
...
+creation of collection products by recursively searching for all label files within a
+specified directory and extracting their identifying information. This information is then
+compared against user-provided details (the bundle and collection names) to determine
+which label files are primary members and which are secondary members. The results are
+compiled into a CSV file and placed in the user's specified location, or in the
+collection's location if no specific destination is provided. This allows for a quick and
+efficient way to check the contents of collections.
+
```
rfrenchseti, line 17: Say something about us not providing label files for the collection product, and the provider needing to work with their assigned Node to develop an appropriate label.

pds4indextools/pds4_create_collection_index.py
```diff
@@ -0,0 +1,81 @@
...
+
+    args = parser.parse_args()
+
+    collection_path = Path(args.collectionpath)
+    label_files = []
+    primary = f"{args.bundle}:{args.collection}"
+
+    # Gather all XML files in the collection path, sort by directory structure
+    label_files = collection_path.glob('**/*.xml')
+    label_files = sorted(label_files, key=lambda x: x.parent)
```
rfrenchseti, line 42: Just to be clear, this sort does automatically exhaust the iterator, meaning the glob will run to completion. For a large directory structure this could take some time. But if you want to sort on directory/filename, there really isn't any other choice.

I'm a little confused by the sort key, though. Isn't the parent the directory containing the file? So what happens to the sort order within a particular directory. It seems like that is unconstrained and will end up being random.

rfrenchseti (changes requested):
- I don't think you fully tested the new documentation. First of all you have to add the new RST file to `index.rst` in order for it to show up at all. Then when you build the docs you'll see warnings about your new file.
- You need to add this tool to the description in the README file. List it as a tool available, and then add a "Getting Started" section for it.
- I'm not fond of having three positional arguments with no names in front of them. It's too easy to forget which is the directory, the bundle, and the collection. Also Matt had a good idea about allowing defaults for these arguments if they aren't specified, and you can't do that if they are required positional arguments. I would go back to having `--bundle` and `--collection` but allow them to be optional arguments, and if they aren't present use the defaults as described by Matt on Slack (collection is the last directory name, and bundle is the directory name above it). First convert the directory name to its absolute path form so things like `.` work properly. Then have the same kind of messages output as we do in the other tool:
```
Creating collection product for: <dir>
Bundle name: <bundle>
Collection name: <collection>
Writing out to: <file>
```
- Are there no errors that can happen anywhere in the code? It seems like maybe we need the occasional try/except block for something, but maybe not.

README.md
```diff
@@ -33,6 +33,8 @@ The following tools are currently available:
 - [`pds4_create_xml_index`](https://rms-pds4indextools.readthedocs.io/en/latest/pds4_create_xml_index.html):
   A command-line program to scrape information from a series of PDS4 XML labels, usually
   in a single collection, and generate a summary index file.
+- [`pds4_create_collection_index`](https://rms-pds4indextools.readthedocs.io/en/latest/pds4_create_collection_index.html):
+  A command-line program to create collection products for collections of PDS4 XML labels.
```
rfrenchseti, line 37: Since this program only works on one collection at a time, let's say "to create a collection product from a collection of PDS4 XML labels".

docs/pds4_create_collection_product.rst
```diff
@@ -0,0 +1,60 @@
...
+Required arguments
+^^^^^^^^^^^^^^^^^^
+
+Three command line arguments are required in every run.
+
+The first is the path to the collection you wish to make a collection product for. This
+path will also be referenced as the output path for the collection product if it is not
+otherwise specified.
+
+The second is the name of the bundle.
```
rfrenchseti, line 38: This needs to be updated with the `--bundle` and `--collection` arguments, and put them below under the optional arguments. There is only one required argument.

pds4indextools/pds4_create_collection_index.py
```diff
@@ -0,0 +1,106 @@
...
+    else:
+        collection = collection_path.name
+
+    print(f"Bundle name: {bundle}")
+    print(f"Collection name: {collection}")
+
+    primary = f"{bundle}:{collection}"
+
+    # Gather all XML files in the collection path, sort by directory structure
+    label_files = collection_path.glob('**/*.xml')
```
rfrenchseti, line 57: As discussed, this needs to be `iglob` and move the if below to later.

pds4indextools/pds4_create_collection_index.py
```diff
@@ -0,0 +1,106 @@
...
+
+        lidvid = f"{lid}::{vid}"
+
+        # Determine member status
+        member_status = 'P' if primary in lid else 'S'
+
+        # Append the data to the list
+        data.append([member_status, lidvid])
+
+    # Sorting the list of results
```
rfrenchseti, line 90: Put the if checking for no label files here (check the data list)

pds4indextools/pds4_create_collection_index.py
```diff
@@ -0,0 +1,106 @@
...
+        tree = etree.parse(str(label_file))
+        root = tree.getroot()
+
+        # Handle namespaces
+        namespaces = root.nsmap
+        namespaces['pds'] = namespaces.pop(None)
+
+        # Extract LID and VID
+        try:
+            lid = tree.find('.//pds:logical_identifier', namespaces=namespaces).text
```
rfrenchseti, line 76: Let's split these into two try/except blocks and for each make the error message more explicit say what explicitly is missing from the label

pds4indextools/pds4_create_collection_index.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # Sorting the list of results
+    data.sort()
+    # Determine the output file path
+    collprod_filepath = (
+        args.collection_product_file if args.collection_product_file else
+        collection_path / f'collection_{args.collection}.csv')
+
+    # Write data to CSV file
+    print(f"Writing out to: {collprod_filepath}")
```
rfrenchseti, line 98: Writing collection product to:

rfrenchseti (changes requested):

pds4indextools/pds4_create_collection_index.py
esimpsons3ti: This uses pathlib.Path.glob, which returns a generator object, not a list of files.

rfrenchseti: Oooo nice catch.

README.md
```diff
@@ -93,8 +98,8 @@ The simplest use scrapes all XML labels from a collection and generates a collec
 pds4_create_collection_product <collection_dir> --bundle bundle_name --collection collection_name
 ```
 
-This will generate a collection product in the given collection directory. Full
-instructions on usage can be found here:
+This will generate a collection product call ``collection_<collection_name>.csv`` in the
```
esimpsons3ti, line 101: call --> called

rfrenchseti (approved):

PR #45, opened 2025-02-11
New functionality: `--replace-headers`

### LINKED ISSUE #21: File to allow renaming of headers
In `pds4_create_xml_index`, right now a header can be one of two strings:

- The full XPath
- The simplified XPath, assuming it is unambiguous

There are use cases where neither of these is desirable, such when XPaths are ambiguous but you don't want to include the full, long XPaths.

One solution is to allow an optional file for the renaming of XPaths. Its format would be something like:

```
<XPath>,<new name>
```

esimpsons3ti:
This pull request will introduce the new feature, `--replace-headers`. This command line argument allows the user to change the names of columns via a user-provided text file. For example, the contents of this file can be:

```
pds:Product_Observational/pds:Identification_Area<1>/pds:logical_identifier<1>,LID
pds:Product_Observational/pds:Identification_Area<1>/pds:version_id<1>,VID
pds:Product_Observational/pds:Identification_Area<1>/pds:title<1>,Title
```

and it will replace the XPath with the given replacement string. This command can be applied to both index files and header files. 

Fixes #21


rfrenchseti (changes requested):
This seems to do what I was hoping for. One small thing is if you have a blank line in your mapping file, it reports it as an error. It would be nice to skip over blank lines so people can organize their field renaming with gaps between them. As an added feature it would be nice to allow comment lines starting with "#" that are ignored. This already works for limit-xpaths files, probably just because there will never be an XML tag that starts with "#" so it never matches.

rfrenchseti (approved):

PR #46, opened 2025-02-18
New feature: `--dont-number-unique-tags`

### LINKED ISSUE #29: New command line argument to remove numbers for disambiguated XPaths
In `pds4_create_xml_index` there are currently two options for the headers:

- Full XPaths include `<1>` numbering at every level
- Disambiguated XPaths where the entire hierarchy is removed except for the final tag whenever possible

I would like to see a new argument called something like `--dont-number-unique-tags` that still gives full XPaths, but if there is only a single instance at a given level, it doesn't include the `<1>` at the end. So instead of:

```
a<1>/b<1>/c<1>
a<1>/b<1>/c<2>
a<1>/b<2>/c<1>
a<1>/b<2>/c<2>
```

we would get:

```
a/b<1>/c<1>
a/b<1>/c<2>
a/b<2>/c<1>
a/b<2>/c<2>
```

This is a very simple modification to the existing renumbering function, where if the mapping only has one value in it we don't add the `<n>` at all.

This function could be used in addition to `--simplify-xpaths` to get the simplest version possible.

esimpsons3ti:
This pull request will introduce a new command-line argument: `--dont-number-unique-tags`. When used, this command will remove the predicates from commonly-shared elements within full or simplified XPaths. When combined with `--simplify-xpaths`, it will create the shortest results for headers.

For example, this command will turn:
```
../pds:logical_identifier<1>
../pds:version_id<1>
../pds:keyword<1>
../pds:keyword<2>
../pds:keyword<3>
```

into:

```
../pds:logical_identifier
../pds:version_id
../pds:keyword<1>
../pds:keyword<2>
../pds:keyword<3>
```

Includes 100% code coverage and unit tests.

Fixes #29


pds4indextools/pds4_create_xml_index.py
```diff
@@ -1311,6 +1318,13 @@ def main(cmd_line=None):
         help='If specified, only writes the tags of unique XPaths to output files. Any '
              'values with duplicate values will still use their full XPath.')
 
+    index_file_generation.add_argument('--dont-number-unique-tags', action='store_true',
+                                       help='If specified, only retain predicates for '
+                                            'elements in XPaths that are unique to their '
+                                            'hierarchy. If multiple instances of an '
+                                            'element exists, the predicates will be '
```
rfrenchseti, line 1325: exists -> exist

rfrenchseti (changes requested):

rfrenchseti (approved):

PR #51, opened 2025-08-20
Fixed compatibility of `--dont-number-unique-tags` and `--limit-xpaths-file`, fixed `--rename-headers` crash.

### LINKED ISSUE #48: pds4_create_xml_index crashes when renaming fields
If you rename an "extra file info" field like this:

```
lid,LID
bundle,bundle_name
```

using the `--replace-headers` option, then the program will crash with:

```
  File "/seti/opus/venv_opus_3.12/bin/pds4_create_xml_index", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/seti/all_repos/rms-pds4indextools/pds4indextools/pds4_create_xml_index.py", line 1689, in main
    true_type = true_type.split(':')[-1]
                ^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'split'
```

This happens because of these lines:

```
                if (header in valid_add_extra_file_info and 'lid' in header):
                    true_type = 'pds:ASCII_LID'
                elif header == 'filename':
                    true_type = 'pds:ASCII_File_Name'
                elif header == 'filepath':
                    true_type = 'pds:ASCII_File_Specification_Name'
                elif header == 'bundle':
                    true_type = 'pds:ASCII_Text_Preserved'
```

When the header has been renamed, it no longer satisfies `header in valid_add_extra_file_info`. Also I think the `if` structure in general is wrong. It should be:

```
                if header in valid_add_extra_file_info:
                    if 'lid' in header:
                        true_type = 'pds:ASCII_LID'
                    elif header == 'filename':
                        true_type = 'pds:ASCII_File_Name'
                    elif header == 'filepath':
                        true_type = 'pds:ASCII_File_Specification_Name'
                    elif header == 'bundle':
                        true_type = 'pds:ASCII_Text_Preserved'
```

### LINKED ISSUE #49: `--limit-xpaths-file` does not play well with `--dont-number-unique-tags`
The help entry for `--limit-xpaths-file` says:

```
--limit-xpaths-file XPATHS_FILEPATH: Specify a text file containing a list of specific XPaths to extract from the label files. If this argument is not specified, all elements found in the label files will be included. This command uses only the whole versions of the XPath(s) – simplified versions are not allowed. 
```

The fact that it wants the full XPath is true in the context of `--simiplify-xpaths`. However, it is wrong for `--dont-number-unique-tags`. When you specify the latter, the XPath file is expecting full XPaths, but without the identifying numbers, like this:

```
pds:Product_Observational/pds:Identification_Area/pds:version_id
```

Either this should be fixed in the code or the documentation should be updated to give this interpretation. Personally I think that `--limit-xpaths-file` should always take the full XPath, including the numbers, and removing the numbers should be part of the output process.

### LINKED ISSUE #50: pds4_create_xml_index: Inconsistencies between documentation and implementation
- The documentation says that `--generate-label` can produce `Product_Ancillary` or `Product_Metadata_Supplemental` by using the `ancillary` or `supplemental` arguments. However, the actual code supports `ancillary` and `metadata`.

- The documentation references the parameter `--label-user-input` for customizing labels, but no such argument exists.

esimpsons3ti:
There is a current problem where, when `--dont-number-unique-tags` and `--limit-xpaths-file` are used together, the latter's input file MUST reflect the format of the cleaned XPaths. The intended behavior should be that the --limit-xpaths-file file can contain the whole XPaths (containing all the predicates) and the command for `--dont-number-unique-tags` should still work. 

Summary of How This Was Fixed:

1. The functionality for removing predicates in `renumber_functions` has been removed, so that all renumbering can be done without removing the predicates.
2. A new function, `clean_predicates` has been introduced. This isolates the behavior for cleaning the predicates.
3. The filtering of the XPaths now happens first, before `--dont-number-unique-tags` is checked.

In addition, two errors in the documentation have been corrected ("supplemental" is actually "metadata", and I removed the mention of a nonexistent `--label-user-input` command.

**New Fix (8/27): `--rename-headers` no longer crashes when used with label generation**
There was a bug where if a column added with `--add-extra-file-info` was renamed using `--rename-headers`, the label generator would search for the original names of these terms. This has now been fixed.

Fixes #48 
Fixes #49 
Fixes #50 
 
 <div id='description'>
    <a href="https://bito.ai#summarystart"></a>
<h3>Summary by Bito</h3>
This pull request addresses compatibility issues between --dont-number-unique-tags and --limit-xpaths-file, updating renumbering logic by separating predicate cleaning into a dedicated function. Documentation was updated to reflect accurate parameter usage and remove unsupported command references, ensuring proper label file processing and consistent XPath handling.

</div>

## Summary by CodeRabbit

- New Features
  - Consistent header renaming via user-provided mapping with stricter validation (invalid or duplicate mappings now produce explicit errors).
  - Improved handling of predicate/numbering edge cases for clearer final headers.

- Documentation
  - Expanded guidance on the exact processing order for header transforms, interaction of options, and how output-headers-file reflects final transformed names.
  - Clarified that sorting and related transforms operate on transformed header names.

- CLI
  - --generate-label now accepts {ancillary, metadata}.

- Tests
  - New end-to-end tests and sample mapping files covering rename-header scenarios.


rfrenchseti (changes requested):
- This does work, but there's one little thing I'd like to see added to the documentation and the usage strings to make this less confusing. Right now it's not at all clear what order the various operations go in. I think somewhere it would be good to mention the explicit order:

```
limit-xpaths => add-extra-file-info => dont-number-unique => simplify-xpaths => rename-headers
```

(assuming I got that right). Right now it's not clear whether the names you give in the rename file are simplified or unnumbered or whatever. It would be good if each command line option mentioned where it happens in the sequence, and the main documentation updated to match along perhaps with a brief sentence or two explaining this.

- Does the code at lines 1620-1624 (for `if args.simplify_xpaths`) actually do anything? `original_headers` doesn't seem like it's ever used again. Perhaps this is old dead code?

- The docstring for `renumber_xpaths` still talks about `dont_number_unique_tags`.

- Not necessary to fix for this PR, but it seems like there's a lot of duplicate code between lines 1635-1680 and the function `write_results_to_csv`. Since what you want is the headers that would be written to the CSV file, wouldn't it be easier to just put the writing of the headers in the write function? There could be a parameter to select between writing a real CSV and a headers file, and then at the end of that function, when you are ready to write the CSV file (and thus know all the real header names), you could write the header text file instead. This would prevent future bugs because right now there are two sequences of code where the various simplify routines are invoked, and if one got changed in the future the other might be overlooked.

rfrenchseti (changes requested):
- It appears that the rename file needs to have an entry for every single column. I would want it to only include the fields that we don't already like the name of. For example, when I was using a rename file for a handful of columns, but not the column `pds:version_id`, I got the following error:

```
  File "/seti/all_repos/rms-pds4indextools/venv/bin/pds4_create_xml_index", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/seti/all_repos/rms-pds4indextools/pds4indextools/pds4_create_xml_index.py", line 1733, in main
    or (mapping and mapping[header] in valid_add_extra_file_info)
                    ~~~~~~~^^^^^^^^
KeyError: 'pds:version_id'
```

- Once that issue is fixed, the rename of headers still doesn't work. The type lookup fails. See code in-line comment.

test_files/samples/rename_headers_label.txt
```diff
@@ -0,0 +1,5 @@
+lid,LID
+bundle,bundle_name
+bundle_lid,bundle_LID
+filename,file_name
+filepath,path_to_file
```
rfrenchseti, line 5: The reason my problem is not being caused by the test is that you're only testing the added extra fields, not the normal fields taken from labels where type lookup is required.

rfrenchseti (changes requested):
- Also, not a big deal, but it seems odd that `--sort-by` is in the middle of the name transformations, before the final rename. Perhaps we can address that in a future PR.

rfrenchseti:
@coderabbitai full review

rfrenchseti (approved):
I'm going to go ahead and approve and merge this PR because it includes some changes necessary to make things work.

